PR #14, opened 2023-08-25
Profits refactor

bergarces:


src/adapters/stargate/products/pool/stargatePoolAdapter.ts
```diff
@@ -84,78 +86,68 @@ export class StargatePoolAdapter implements IProtocolAdapter {
...
           balanceRaw: BigInt(amountLPtoLD.toString()),
           balance: formatUnits(amountLPtoLD, underlyingTokenMetadata.decimals),
           type: TokenType.Underlying,
         }
+
+        return {
+          ...protocolTokenBalance,
+          type: TokenType.Protocol,
+          tokens: [underlyingTokenBalance],
+        }
```
bergarces, line 109: Moved this inside the map to avoid unchecked index access.

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
```diff
@@ -84,78 +86,68 @@ export class StargatePoolAdapter implements IProtocolAdapter {
...
-    const underlying = await Promise.all(
+    const tokens = await Promise.all(
       protocolTokensBalances.map(async (protocolTokenBalance) => {
         const amountLPtoLD = await this.stargateTokenContract(
           protocolTokenBalance.address,
         ).amountLPtoLD(protocolTokenBalance.balanceRaw)
 
         const underlyingTokenMetadata =
-          this.metadata[protocolTokenBalance.address].underlying
+          this.metadata[protocolTokenBalance.address]!.underlying
```
bergarces, line 96: I can't see how to avoid the bang here without creating a lot of messy code. We know `getBalances` returns a balance for each lp in metadata, and we know that every metadata pool has one and only one underlying token.

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
```diff
@@ -216,76 +214,76 @@ export class StargatePoolAdapter implements IProtocolAdapter {
...
+                  type: TokenType.Underlying,
+                }
+              },
+            ),
+          }
+        },
+      ),
     )
 
-    const tokens = getProfitPerPosition.map((values) => {
```
bergarces, line 255: Moved this block to previous map to avoid indexed accesses.

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
jpsains: So getBalance also filters out 0 balances, which not an issue, as long as getBalance doesn't do anything funky like changing the lp address from '0x0000' to '0xpotatoe' then we should be ok

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
```diff
@@ -216,76 +214,84 @@ export class StargatePoolAdapter implements IProtocolAdapter {
...
         userAddress,
         blockNumber: toBlock,
-      }).then((result) => formatToMap(result)),
-
+      }).then((positions) => formatProtocolTokenArrayToMap(positions)),
       this.getPositions({
         userAddress,
         blockNumber: fromBlock,
-      }).then((result) => formatToMap(result)),
+      }).then((positions) => formatProtocolTokenArrayToMap(positions)),
```
jpsains, line 221: image

nit picking but might be a nice iteration

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
```diff
@@ -216,76 +214,84 @@ export class StargatePoolAdapter implements IProtocolAdapter {
...
-          }).then((result) => aggregateTrades(result)),
-        ])
-
-        return {
-          [protocolTokenAddress]: calculateProfit(
+          }
+
+          const [withdrawal, deposits] = await Promise.all([
+            this.getWithdrawals(getEventsInput).then((movements) =>
+              aggregateTrades(movements),
```
jpsains, line 236: same here

src/core/utils/calculateProfit.ts
```diff
@@ -1,23 +1,28 @@
+import { BaseToken } from '../../types/adapter'
+
 export function calculateProfit(
-  deposits: Record<string, number | undefined> | undefined,
-  withdrawals: Record<string, number | undefined> | undefined,
-  currentValues: Record<string, number | undefined> | undefined,
-  previousVales: Record<string, number | undefined> | undefined,
-): Record<string, number> {
+  deposits: Record<string, bigint>,
```
jpsains, line 4: can you right click on calculateProfit(...) then click refactor, then convert this to accept an object please (I know it was me that implemented but its confusing)

jpsains (approved):

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
bergarces: Done. I'm always weary of using that for readability, but in this case it's pretty obvious.

src/core/utils/calculateProfit.ts
bergarces: Done.

src/adapters/stargate/products/pool/stargatePoolAdapter.ts
bergarces: Yes, as long as `getBalances` doesn't return anything that wasn't already on the list, we are good. It's a `utils` function, so it should not be normally changed by external contributors.

jpsains (approved):

PR #19, opened 2023-08-30
MulticallQueue interceptor

jpsains:
MulticallQueue interceptor

- To activate set ENABLE_MULTICALL to true. Set as env-var so that we have a kill switch.
- This then routes all eth_calls to a multicall smart contract.
- We can configure batch size and flush timeout.
- DID YOU SEE INCEPTION? Yes when ENABLE_MULTICALL=true our balanceOf multicall request is routed via the new MulticallQueue. Therefore a multicall contract is calling a different multicall contract. It works and I dont see an issue atm with keeping it that way. I expect this is fairly common thing, e.g. Uniswap has its own multicall function which Im sure is called often by other multicalls


src/core/utils/multicall.ts
```diff
@@ -0,0 +1,108 @@
...
+          callsToProcessLength,
+        },
+        'Multicall response length differs from batch sent',
+      )
+      return
+    }
+
+    callsToProcess.forEach(({ resolve, reject }, i) => {
+      const returnedData = results[i]?.returnData
+      const success = results[i]?.success
```
jpsains, line 99: Regarding '?' I am not totally sure why it's not happy here. Might be something to do with the Multicall contract interface. I feel like I am covering issues with  `(resultLength !== callsToProcessLength)` check but happy to be proven wrong

src/core/utils/chainProviders.ts
```diff
@@ -1,23 +1,104 @@
...
+  enableMulticall,
+}: {
+  url: string | undefined
+  chainId: Chain
+  enableMulticall: boolean
+}) => {
+  logger.info('Using multicall queue providers')
+  if (!url) return undefined
+
+  if (enableMulticall) {
```
bergarces, line 23: If multicall is set globally, we can just check `process.env.ENABLE_MULTICALL === 'true'` here and avoid having to pass the same parameter for every chain.

src/core/utils/chainProviders.ts
```diff
@@ -1,23 +1,104 @@
...
+
+    return new CustomMulticallJsonRpcProvider({
+      url,
+      network: chainId,
+      multicallQueue,
+    })
+  }
+
+  logger.info('Using standard providers')
+  return new ethers.providers.StaticJsonRpcProvider(url, chainId)
```
bergarces, line 50: nit: Since this is the short path, it'd be nicer if the branching was the other way around:
```
if (!enableMulticall) {
  logger.info('Using standard providers')
  return new ethers.providers.StaticJsonRpcProvider(url, chainId)
}

// Longer chunk of code to deal with the multicall approach.
```

src/core/utils/multicall.ts
bergarces: With `"noUncheckedIndexedAccess": true,`, any access via index will always return `T | undefined`.

If you know for sure from your internal implementation that it cannot be undefined because you have checked it beforehand, but TypeScript cannot parse it properly, then you need to use `results[i]!.success`.

src/core/utils/chainProviders.ts
jpsains: My thinking is some chains might not have the multicall deployed yet

jpsains: so we have some control at a chain level

src/core/utils/chainProviders.ts
jpsains: Yep agreed

.env.example
```diff
@@ -10,3 +10,5 @@ BASE_PROVIDER_URL=https://1rpc.io/base
 ARBITRUM_PROVIDER_URL=https://arbitrum-mainnet.infura.io/v3/abafec30d6aa45ffa0c763b5552a2d02
 AVALANCHE_PROVIDER_URL=https://avalanche-mainnet.infura.io/v3/abafec30d6aa45ffa0c763b5552a2d02
 LINEA_PROVIDER_URL=https://linea-mainnet.infura.io/v3/abafec30d6aa45ffa0c763b5552a2d02
+
+ENABLE_MULTICALL=false
```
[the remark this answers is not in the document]
jpsains: might call it ENABLE_MULTCALL queue

src/core/utils/CustomMulticallJsonRpcProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+  }) {
+    super(url, network)
+    this.multicallQueue = multicallQueue
+  }
+
+  async call(
+    transaction: providers.TransactionRequest,
+    blockTag?: providers.BlockTag | Promise<providers.BlockTag>,
+  ): Promise<string> {
+    console.log('Intercepted eth_call')
```
bergarces, line 23: I'd suggest `logger.debug` here, it might be a bit spammy otherwise.

src/core/utils/CustomMulticallJsonRpcProvider.test.ts
bergarces, line 1: nit: We haven't discussed unit test structure, but just so that they are easier to follow I'd suggest the following:

Top: `describe('CustomMultiCallProvider')`
For each method: `describe('methodName')` (I'd avoid testing the constructor unless there's logic, if it's just setup it'll get covered with the regular tests)
For each test: `it('<what's being tested>')` (present tense, like you've done)

Then, for bigger classes with more tests, we might want to declare the instance of the class and spies within the `describe` and populate them with `beforeEach` or `beforeAll`, but in this case it's not worth doing.

bergarces (approved):
All good. None of the comments is a blocker.

src/core/utils/multicall.ts
jpsains: ok played alot with this and managed to get ts to stop complaining a simple if(!results[i]) didn't do the trick but assigning the value to a new var and checking that then stopped ts complaining 

```     
 const result = results[i]
      if (!result) return
```

src/core/utils/CustomMulticallJsonRpcProvider.test.ts
jpsains: legendary catch

src/core/utils/CustomMulticallJsonRpcProvider.ts
jpsains: nice idea

PR #32, opened 2023-09-08
Reorganize metadata

bergarces:
- Remvoes the need for `buildMetadata` methods for each protocol by adding `buildMetadata` as part of the adapter.
- `src/adapters/index.ts` does no longer need to include a factory method, just a list of adapters.
- Adds decorators to allow the previous point
- Fixes existing templates


package.json
```diff
@@ -66,6 +66,7 @@
     "recast": "^0.23.4",
     "ts-jest": "^29.1.1",
     "ts-node": "^10.9.1",
+    "tsc-hooks": "^1.1.2",
```
bergarces, line 69: `tsc` does not copy any metadata file that is not directly referenced as an `import`, this package helps with that.

jpsains: 👏

PR #49, opened 2023-09-14
add stargate contracts to adapter folder

bergarces:


src/core/utils/getAddressesBalances.ts
```diff
@@ -34,11 +41,27 @@ export async function getAddressesBalances({
...
+  values: BigNumber[],
+  addresses: string[],
+  tokens: string[],
+): AddressBalanceMap {
+  const balances: AddressBalanceMap = {}
+  addresses.forEach((addr, addrIdx) => {
+    balances[addr] = {}
+    tokens.forEach((tokenAddr, tokenIdx) => {
+      const balance = values[addrIdx * tokens.length + tokenIdx]
+      balances[addr]![tokenAddr] = balance!.toString()
```
jpsains, line 63: can the balance request ever fail?

What happens when one fails? does the whole thing fail?

jpsains (approved):

src/core/utils/getAddressesBalances.ts
bergarces: Yes. If getting the balance for any token fails, the whole contract call fails (https://etherscan.io/address/0xb1f8e55c7f64d203c1400b9d8555d050f94adf39#code).

I copied this function from their library and just added the typescript bits.

PR #50, opened 2023-09-15
Ethers v6

bergarces:
Upgrade to ethersv6, no more `BigNumber`.


src/core/utils/multicall.ts
```diff
@@ -81,7 +85,7 @@ export class MulticallQueue {
     const batchSize = callsToProcess.length
     logger.debug({ batchSize }, 'Sending multicall batch ')
 
-    const results = await this.multicallContract.callStatic.aggregate3(
```
bergarces, line 84: They changed the way this is called.

src/core/utils/CustomMulticallJsonRpcProvider.ts
```diff
@@ -1,33 +1,31 @@
...
-    if (blockTag) {
-      logger.debug('Intercepted eth_call, using multicall queue implementation')
-      return super.call(transaction, blockTag)
+  async call(transaction: TransactionRequest): Promise<string> {
+    logger.debug(
+      transaction,
+      'Intercepted eth_call, using multicall queue implementation',
+    )
+    if (transaction.blockTag) {
+      return super.call(transaction)
```
bergarces, line 26: `blockTag` is now part of the `TransactionRequest`, they split it out internally within the provider.

src/core/utils/chainProviders.ts
```diff
@@ -20,20 +19,21 @@ const provider = ({
 
   if (!enableMulticallQueue) {
     logger.debug({ chainId }, `Using standard provider`)
-    return new ethers.providers.StaticJsonRpcProvider(url, chainId)
+    return new ethers.JsonRpcProvider(url, chainId, {
+      staticNetwork: Network.from(chainId),
+    })
```
bergarces, line 24: There's no `StaticJsonRpcProvider`, the new way of using it is by passing a `staticNetwork` property to the settings.

src/core/utils/chainProviders.ts
```diff
@@ -20,20 +19,21 @@ const provider = ({
...
 
   logger.debug({ chainId }, 'Using multicall queue provider')
 
-  const provider = new ethers.providers.StaticJsonRpcProvider(url, chainId)
+  const provider = new ethers.JsonRpcProvider(url, chainId)
 
   // deployed on 100+ chains at address
   // https://www.multicall3.com/deployments
-  const multicallContract = new Contract(
+  const multicallContract = Multicall__factory.connect(
```
bergarces, line 33: Is there any reason why we were using `ethers.Contract` directly here? It's throwing type errors now.

src/core/utils/chainProviders.ts
jpsains: makes sense

src/core/utils/chainProviders.ts
jpsains: It is fine like you proposed, the thinking was MulticallQueue would have one less dependency in case we wanted to separate to own library, but its fine what you have.

src/core/utils/multicall.ts
jpsains: Ok good, in solidity I think they call it `staticcall` I'm glad ethers changed

src/core/utils/CustomMulticallJsonRpcProvider.ts
jpsains: oh thats cool are there any overrides now?

jpsains (approved):

src/core/utils/CustomMulticallJsonRpcProvider.ts
bergarces: These are all the overrides. Before, `blockTag` was an additional optional parameter. Much better this way IMO.

https://github.com/ethers-io/ethers.js/blob/main/src.ts/providers/provider.ts#L124

PR #53, opened 2023-09-18
feat: apollo graphql setup

aganglada:


jpsains:
Thanks for this Alej, you've planted the seed, will circle back to this if we get some requests for a UI

PR #55, opened 2023-09-19
feat: profits improvements

jpsains:
Changes 

1. Introduced "Lend' position type so that we can distinguish between a lending position such as Aave and lp position such as uniswap which we currently call "Supply".  
2. Update profits to be more user-friendly

### DIFF
--- a/src/core/adapters/SimplePoolAdapter.ts
+++ b/src/core/adapters/SimplePoolAdapter.ts
@@ -140,13 +140,11 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
     input: GetTotalValueLockedInput,
   ): Promise<ProtocolTotalValueLockedToken[]>
 
-  async getOneDayProfit({
+  async getProfits({
     userAddress,
-    blockNumber,
+    toBlock,
+    fromBlock,
   }: GetProfitsInput): Promise<ProfitsTokensWithRange> {
-    const toBlock = blockNumber
-    const fromBlock = toBlock - AVERAGE_BLOCKS_PER_DAY[this.chainId]
-
     const [currentValues, previousValues] = await Promise.all([
       this.getPositions({
         userAddress,
--- a/src/core/adapters/SimplePoolAdapter.ts
+++ b/src/core/adapters/SimplePoolAdapter.ts
@@ -27,7 +27,6 @@ import {
   TokenBalance,
   TokenType,
 } from '../../types/adapter'
-import { AVERAGE_BLOCKS_PER_DAY } from '../constants/AVERAGE_BLOCKS_PER_DAY'
 import { Chain } from '../constants/chains'
 import { ZERO_ADDRESS } from '../constants/ZERO_ADDRESS'
 import { aggregateTrades } from '../utils/aggregateTrades'
--- a/src/core/adapters/SimplePoolAdapter.ts
+++ b/src/core/adapters/SimplePoolAdapter.ts
@@ -141,8 +141,8 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
 
   async getProfits({
     userAddress,
-    toBlock,
     fromBlock,
+    toBlock,
   }: GetProfitsInput): Promise<ProfitsTokensWithRange> {
     const [currentValues, previousValues] = await Promise.all([
       this.getPositions({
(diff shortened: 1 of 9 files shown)


src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -140,13 +139,11 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
...
-  async getOneDayProfit({
+  async getProfits({
     userAddress,
-    blockNumber,
+    toBlock,
+    fromBlock,
   }: GetProfitsInput): Promise<ProfitsTokensWithRange> {
-    const toBlock = blockNumber
-    const fromBlock = toBlock - AVERAGE_BLOCKS_PER_DAY[this.chainId]
-
```
jpsains, line 149: block numbers now gets passed into function which feels cleaner

bergarces: This is going to be a massive benefit for the integration tests too as  it will allow consistency even if we update the average blocks per month. I like it.

src/index.ts
```diff
@@ -15,18 +17,33 @@ import {
...
 export async function getPositions({
   userAddress,
   filterProtocolIds,
   filterChainIds,
+  blockNumbers,
 }: {
   userAddress: string
   filterProtocolIds?: Protocol[]
   filterChainIds?: Chain[]
+  blockNumbers?: Record<string, number | undefined>
```
bergarces, line 37:
```suggestion
  blockNumbers?: Partial<Record<Chain, number>>
```
This will achieve the same and prevent using keys that don't exist.

src/index.ts
```diff
@@ -15,18 +17,33 @@ import {
...
 }): Promise<DefiPositionResponse[]> {
   const runner = async (adapter: IProtocolAdapter) => {
+    const chainId = adapter.getProtocolDetails().chainId
```
bergarces, line 40:
```suggestion
    const chainId = adapter.chainId
```

`chainId` is a property of `IProtocolAdapter`.

src/index.ts
```diff
@@ -39,23 +56,33 @@ export async function getPositions({
...
   filterProtocolIds,
   filterChainIds,
+  toBlockNumbersOverride,
+  filterTimePeriod = TimePeriod.sevenDays,
 }: {
   userAddress: string
+  filterTimePeriod?: TimePeriod
   filterProtocolIds?: Protocol[]
   filterChainIds?: Chain[]
+  toBlockNumbersOverride?: Record<Chain, number | undefined>
```
bergarces, line 71:
```suggestion
  toBlockNumbersOverride?: Partial<Record<Chain, number>>
```

src/index.ts
```diff
@@ -39,23 +56,33 @@ export async function getPositions({
...
 }): Promise<DefiProfitsResponse[]> {
   const runner = async (
     adapter: IProtocolAdapter,
     provider: ethers.JsonRpcProvider,
   ) => {
-    const blockNumber = await provider.getBlockNumber()
-    return await adapter.getOneDayProfit({
+    const chainId = adapter.getProtocolDetails().chainId
```
bergarces, line 77:
```suggestion
    const chainId = adapter.chainId
```

src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -140,13 +139,11 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
     input: GetTotalValueLockedInput,
   ): Promise<ProtocolTotalValueLockedToken[]>
 
-  async getOneDayProfit({
+  async getProfits({
     userAddress,
-    blockNumber,
+    toBlock,
+    fromBlock,
```
bergarces, line 145: nit: Change the order of these two :sweat_smile:

src/index.ts
```diff
@@ -39,23 +56,33 @@ export async function getPositions({
...
+
+    const toBlock =
+      toBlockNumbersOverride?.[chainId] || (await provider.getBlockNumber())
```
bergarces, line 80:
```suggestion
      toBlockNumbersOverride?.[chainId] ?? (await provider.getBlockNumber())
```
The null coalescing operator is `??`. If your block override was `0`, with `||` it'd be considered invalid and it'll get a new blockNumber.

Ok, I know that chances are the `toBlock` won't be 0, but as a rule of thumb, when you want to do something if the left expression is null or undefined, the correct operator is `??`.

bergarces (approved):

PR #57, opened 2023-09-20
feat: borrow token profit support

jpsains:


src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -257,33 +268,44 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
...
+                throw new Error('No price for underlying token at this time')
+              }
+
+              const protocolTokenMovementValue = formatUnits(
+                protocolTokenMovementValueRaw,
+                protocolTokenMetadata.decimals,
+              )
+
+              const movementValue =
+                +protocolTokenMovementValue * +currentTokenPrice.pricePerShare
```
jpsains, line 296: These are both decimal values. I think our formula and approach could do with a specification or mini white-paper to understand the implications fully, loss of precision is an acceptable risk, these are read-only values which will get converted into USD.

src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -257,33 +268,44 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
...
-          throw new Error('No price for events at this time')
-        }
-
-        const movementValueRaw = BigInt(value.toString()) * pricePerShareRaw
         return {
           underlyingTokensMovement: underlyingTokens.reduce(
-            (accummulator, currentToken) => {
+            (accumulator, currentToken) => {
+              const currentTokenPrice = protocolTokenPrice.tokens?.find(
+                (price) => price.address === currentToken.address,
```
bergarces, line 283: Good catch here. We were picking always the first token, but that only works if there's only one underlying.

bergarces (dismissed):
Looks good.

I'd take into consideration if it's worth having values such as `pricePerShare` or `profit` as `string` instead of `number`.

The whole idea of using `string` was to protect the code against values that are too big to fit in a js `number`, but since we are getting them by operating with `number`, that's unnecessary.

src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -184,13 +188,16 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
...
                 return {
                   ...underlyingToken,
-                  profitRaw: profits[underlyingToken.address]!,
-                  profit: formatUnits(
-                    profits[underlyingToken.address]!,
-                    underlyingToken.decimals,
-                  ),
+                  profit: isBorrowPosition
+                    ? (-profit).toString()
+                    : profit.toString(),
```
bergarces, line 200:
```suggestion
                  profit: isBorrowPosition
                    ? (-profit).toString()
                    : profit,
```

`profit` is already a string.

PR #59, opened 2023-09-20
filter 0 values

bergarces:


src/adapters/stargate/products/vesting/stargateVestingAdapter.ts
```diff
@@ -89,6 +89,10 @@ export class StargateVestingAdapter
       blockTag: blockNumber,
     })
 
+    if (!userBalance && !lockedAmount) {
```
bergarces, line 92: I'm pretty sure that one of them should be dominant, but just in case I'm checking both.

jpsains (approved):

PR #71, opened 2023-10-05
Mmi 3924 implement new de fi adapter for uniswap v 3 in new repo

jpsains:


src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,348 @@
...
+            tokenId: tokenId as bigint,
+            liquidity,
+            amount0Min: 0n,
+            amount1Min: 0n,
+            deadline,
+          },
+          { from: userAddress, blockTag: blockNumber },
+        ) as unknown as Promise<
+          [bigint, bigint] & { amount0: bigint; amount1: bigint }
+        >, // IMO typechain has an bug here, incorrect type being returned
```
jpsains, line 172: typechain says the type being returned is a nested array but its not you can check in the function in the abi

Typechain says
[[bigint, bigint] & { amount0: bigint; amount1: bigint }] 

I say 
[bigint, bigint] & { amount0: bigint; amount1: bigint }

abi

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,348 @@
...
+        address: positionManagerCommonAddress,
+        name: nftName,
+        symbol: nftName,
+        decimals: 18,
+        balanceRaw: pos.liquidity,
+        type: TokenType.Protocol,
+        tokens: [
+          this.createUnderlyingToken(
+            pos.token0,
+            erc20Metadata[index]?.[token0Index],
```
bergarces, line 228: If you know it cannot be undefined, then we should be using `erc20Metadata[index]!`

src/adapters/uniswap-v3/contracts/factories/UniswapPoolFactory__factory.ts
bergarces, line 1: Is this contract likely to be needed in future? It doesn't look like it's being used here.

src/adapters/uniswap-v3/contracts/factories/UniswapPool__factory.ts
bergarces, line 1: Is this contract likely to be needed in future? It doesn't look like it's being used here.

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+  ProtocolPosition,
+  TokenType,
+  Underlying,
+} from '../../../../types/adapter'
+import { Erc20Metadata } from '../../../../types/erc20Metadata'
+import { IProtocolAdapter } from '../../../../types/IProtocolAdapter'
+import { Protocol } from '../../../protocols'
+import { PositionManager__factory } from '../../contracts'
+
+const deadline = Math.floor(Date.now() - 1000) + 60 * 10
```
bergarces, line 37: A comment with what this calculation means would be handy.

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  /**
+   * Update me.
+   * Add logic to build protocol token metadata
+   * For context see dashboard example ./dashboard.png
+   * We need protocol token names, decimals, and also linked underlying tokens
+   */
+  @CacheToFile({ fileKey: 'protocol-token' })
+  async buildMetadata() {
```
bergarces, line 100: Is it going to have metadata? If not, we can just not add `IMetadataBuilder` to avoid having to add this method.

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+    throw new Error('Implement me')
+
+    return {}
+  }
+
+  /**
+   * Update me.
+   * Returning an array of your protocol tokens.
+   */
+  async getProtocolTokens(): Promise<Erc20Metadata[]> {
```
bergarces, line 110: We might want to reconsider whether this should be part of the `IProtocolAdapter` public interface.

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,354 @@
...
+      contractAddresses[this.chainId]!.positionManager,
+      this.provider,
+    )
+
+    const balanceOf = await positionsManagerContract.balanceOf(userAddress, {
+      blockTag: blockNumber,
+    })
+
+    const tokenIds = await Promise.all(
+      [...Array(Number(balanceOf)).keys()].map((index) =>
```
bergarces, line 132: This could result in a pretty big loop. We might want to come back to this eventually.

bergarces: I'm unsure what the best policy would be here.

I'm pretty sure we could have a single async map that returns `nonZeroPositions` directly without looping three times with something like this.

```
const nonZeroPositions = await Promise.all(
      filterMap([...Array(Number(balanceOf)).keys()], async (index) => {
        const tokenId = await positionsManagerContract.tokenOfOwnerByIndex(
          userAddress,
          index,
          {
            blockTag: blockNumber,
          },
        )

        const position = await positionsManagerContract.positions(tokenId, {
          blockTag: blockNumber,
        })

        if (position.liquidity == 0n) {
          return undefined
        }

        return {
          liquidity: position.liquidity,
          token0: position.token0,
          token1: position.token1,
          fee: position.fee,
          tokenId: tokenIds[index],
        }
      }),
    )
```

I'm unsure whether it'll save any time at all. We would need to run profiling with users with a lot of positions.

bergarces (dismissed):

src/adapters/uniswap-v3/contracts/factories/UniswapPool__factory.ts
jpsains: ill remove until its needed

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
jpsains: added

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
jpsains: I think it might

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
jpsains: ive implemented

src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -0,0 +1,317 @@
...
+              liquidity: position.liquidity,
+              amount0Min: 0n,
+              amount1Min: 0n,
+              deadline,
+            },
+            { from: userAddress, blockTag: blockNumber },
+          ),
+          positionsManagerContract.collect.staticCall(
+            {
+              tokenId: tokenId as bigint,
```
jpsains, line 170: i should be able to get rid now

PR #73, opened 2023-10-06
aavev2 apr and apy

bergarces:
APR and APY calculation for AaveV2 pools.

Removes multiple frozen and inactive protocol tokens from AaveV2 metadata


src/types/adapter.ts
```diff
@@ -267,14 +267,14 @@ export interface ProtocolTokenApy extends Erc20Metadata {
   /**
    * Current apy of protocol pool
    */
-  apyDecimal: string
+  apyDecimal: number
```
bergarces, line 270: What do we think about this change?

Having it as a string would make it tricky to work with it afterwards.

src/adapters/aave-v2/products/pool/aaveV2BasePoolAdapter.ts
```diff
@@ -31,6 +33,14 @@ type AaveV2PoolMetadata = Record<
   }
 >
 
+const protocolDataProviderContractAddresses: Partial<Record<Chain, string>> = {
```
bergarces, line 36: Same contracts used for metadata, so I've just moved them as a constant for the whole file.

src/adapters/aave-v2/products/pool/aaveV2BasePoolAdapter.ts
```diff
@@ -31,6 +33,14 @@ type AaveV2PoolMetadata = Record<
...
+  [Chain.Ethereum]: '0x057835Ad21a177dbdd3090bB1CAE03EaCF78Fc6d',
+  [Chain.Polygon]: '0x7551b5D2763519d4e37e8B81929D336De671d46d',
+  [Chain.Avalanche]: '0x65285E9dfab318f57051ab2b139ccCf232945451',
+}
+
+const RAY = 10 ** 27
```
bergarces, line 42: According to their glossary, A ray is just a unit with 27 decimals of precision.

https://docs.aave.com/developers/v/2.0/glossary

src/adapters/aave-v2/products/pool/aaveV2BasePoolAdapter.ts
```diff
@@ -87,6 +116,18 @@ export abstract class AaveV2BasePoolAdapter
...
     const metadataObject: AaveV2PoolMetadata = {}
     for (const { tokenAddress } of reserveTokens) {
+      const reserveConfigurationData =
+        await protocolDataProviderContract.getReserveConfigurationData(
+          tokenAddress,
+        )
+
+      if (
+        !reserveConfigurationData.isActive ||
+        reserveConfigurationData.isFrozen
```
bergarces, line 126: Checks that the protocol token is active and not frozen.

src/types/adapter.ts
jpsains: No strong opinion. Dont see an issue atm

src/adapters/aave-v2/tests/snapshots/ethereum.profits.json
```diff
@@ -11,93 +11,6 @@
...
-          "symbol": "aMKR",
-          "decimals": 18,
-          "type": "protocol",
-          "tokens": [
-            {
-              "address": "0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2",
-              "name": "Maker",
-              "symbol": "MKR",
-              "decimals": 18,
-              "profitRaw": "7186772652009n",
```
jpsains, line 26: i think we are missing profit formatted number, but should do on separate mr

jpsains (approved):

src/adapters/aave-v2/tests/snapshots/ethereum.profits.json
bergarces: We are not missing them, they are at the bottom.

They get printed like this because we add them at the end of the process (maybe we can apply some order when printing).

PR #74, opened 2023-10-09
add custom errors and filter when not applicable

bergarces:
Adds custom errors for:
- Method not applicable (which also gets filtered out of console.log output)
- Method not implemented
- Provider not found


src/types/response.ts
```diff
@@ -21,7 +21,8 @@ import { Erc20Metadata } from './erc20Metadata'
 export type AdapterErrorResponse = {
   error: {
     message: string
-    details?: Record<string, unknown>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    details?: any
```
bergarces, line 25: I've done this because in both cases we don't know what this is going to have, but with `any` we can at least check it without typescript complaining.

src/core/errors/errors.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    this.name = 'NotApplicableError'
+  }
+}
+
+export class NotImplementedError extends Error {
+  constructor(message = 'Not Implemented') {
+    super(message)
+
+    Error.captureStackTrace(this, NotImplementedError)
```
jpsains, line 17: this seems like black magic, what's the reason?

jpsains (approved):

src/core/errors/errors.ts
bergarces: That's just to prevent this constructor call from appearing in the stack trace.

It's common place when extending `Error`, but it's not critical.

PR #75, opened 2023-10-09
feat: protocolTokenAddress param

jpsains:


src/index.ts
```diff
@@ -144,93 +145,119 @@ export async function getPrices({
...
+      error: {
+        message: 'No adapter found',
+        details: {
+          product,
+          protocolId,
+          chainId: chainId,
+          chainName: ChainName[chainId],
+        },
+      },
+    } as any
```
bergarces, line 180: I don't think you need casting here. That's a valid return type.

src/core/metadata/AdapterMetadata.ts
```diff
@@ -53,86 +52,78 @@ export const MetadataFiles = new Map<string, Json>([
...
       protocolId: Protocol.AaveV2,
-      productId: 'pool',
+      productId: 'stable-debt-token',
       chainId: Chain.Avalanche,
       fileKey: 'stable-debt-token',
     }),
-    AaveV2PoolAvalancheStableDebtToken,
-  ],
-  [
-    metadataKey({
```
bergarces, line 129: Looks like we are missing this entry.

src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -388,6 +392,12 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
           })
 
         return {
+          protocolToken: {
+            address: protocolToken.address,
+            name: protocolToken.name,
+            symbol: protocolToken.symbol,
+            decimals: protocolToken.decimals,
+          },
```
bergarces, line 400: You can safely just do `protocolToken` here, it only has those fields.

bergarces (approved):
Left some comments, but can be merged and we can deal with them later.

bergarces (approved):
Left some comments, but it can be merged and we can deal with them later.

PR #76, opened 2023-10-11
Mmi 3924 uniswap profits

jpsains:


src/adapters/uniswap-v3/products/pool/uniswapV3PoolAdapter.ts
```diff
@@ -101,7 +101,7 @@ export class UniswapV3PoolAdapter
    */
   @CacheToFile({ fileKey: 'protocol-token' })
   async buildMetadata() {
```
bergarces, line 103: With `IMetadataBuilder` gone. This should go as well.

bergarces (dismissed):
Looks good.

I've taken a couple of low priority notes regarding:
- Most of profit calculation is identical except for the `tokenId` situation. I think something could be done there.
- `tokenId` as an eye-sore necessary exception.

It's not worth investing a lot of time on making some complicated abstraction as long as there's only one exception, so we can wait until another protocol/product adds another optional bit before we suggest another solution.

bergarces:
I've pushed a commit with the conflict resolved.

bergarces (approved):

PR #80, opened 2023-10-17
feat: config

jpsains:


README.md
```diff
@@ -29,20 +29,19 @@ This project requires Node 18. Ensure you're using the correct version (e.g. run
 
 To build an adapter follow these steps:
 
-1. Create a copy of `.env.example` as `.env`.
```
jpsains, line 32: no need for this step now

src/scripts/index.ts
```diff
@@ -10,16 +11,18 @@ import { newAdapterCommand } from './newAdapterCommand'
...
 
 buildContractTypes(program)
 
-buildMetadata(program)
+buildMetadata(program, defiAdapers)
 
-buildSnapshots(program)
+buildSnapshots(program, defiAdapers)
 
 program.parseAsync()
```
jpsains, line 28: Happy to iterate on this

src/scripts/index.ts
```diff
@@ -10,16 +11,18 @@ import { newAdapterCommand } from './newAdapterCommand'
...
-blockAverage(program)
+blockAverage(program, defiAdapers)
 
 buildContractTypes(program)
 
-buildMetadata(program)
+buildMetadata(program, defiAdapers)
 
-buildSnapshots(program)
+buildSnapshots(program, defiAdapers)
```
jpsains, line 27: Here Im passing around one instance, let me know if you have a preferred solution. Injecting class into function might have some benefit in future.

src/config.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.config.validate({ allowed: 'strict' })
+  }
+
+  /**
+   * Get the entire configuration as a typed object.
+   * @returns The configuration object.
+   */
+  get values(): IConfig {
+    return this.config.getProperties()
+  }
```
jpsains, line 128: I could return `this.config` however this implementation looked prettier when accessing the config

```new Config().values.get('provider.polygon')```
```new Config().values.provider.polygon```

File: src/defi-adapters.ts [hunk 5]
```diff
@@ -0,0 +1,470 @@
...
+  DefiPositionResponse,
+  DefiProfitsResponse,
+  PositionType,
+  Protocol,
+  TimePeriod,
+}
+export class DefiAdapter {
+  private config: Config
+  chainProvider: ChainProvider
+
```
jpsains, line 43: Thoughts on not making this private and therefore being able to access?

src/adapters/aave-v2/common/aaveBasePoolAdapter.ts
```diff
@@ -153,10 +153,12 @@ export abstract class AaveBasePoolAdapter
       const protocolTokenPromise = getTokenMetadata(
         this.getReserveTokenAddress(reserveTokenAddresses),
         this.chainId,
+        this.provider,
       )
       const underlyingTokenPromise = getTokenMetadata(
         tokenAddress,
         this.chainId,
+        this.provider,
```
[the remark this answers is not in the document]
bergarces: This should have been like this from the beginning. 100% agree with this change.

src/core/utils/chainProviders.ts
```diff
@@ -1,100 +1,110 @@
...
+  private config: Config
+  providers: Record<Chain, ethers.JsonRpcProvider>
 
-  if (!enableMulticallQueue) {
-    logger.debug({ chainId }, `Using standard provider`)
-    return new ethers.JsonRpcProvider(url, chainId, {
-      staticNetwork: Network.from(chainId),
-    })
+  constructor(config: Config) {
+    this.config = config
```
bergarces, line 15: I don't think you need to store the config. Just pass it to `initializeProviders` so it can be used there.

src/core/utils/chainProviders.ts
```diff
@@ -1,100 +1,110 @@
...
+  private provider({
+    url,
+    chainId,
+    enableMulticallQueue,
+  }: {
+    url: string
+    chainId: Chain
+    enableMulticallQueue: boolean
+  }): ethers.JsonRpcProvider {
+    if (!url) throw new Error('Url missing')
```
bergarces, line 28: I'm going to add `curly: 'error',` as an eslint rule on my next PR.

src/config.ts
bergarces: Agree that this works better.

PR #81, opened 2023-10-18
feat: stableBlockNumber

jpsains:
Extends the ethers.jsonRpcProvider and adds an additional method called getStableBlock.

 Returns 10 min old blockNumber to ensure data has propagated across nodes
 Getting logs close to head has issues

### DIFF
--- a/src/core/constants/AVERAGE_BLOCKS_PER_10_MINS.ts
+++ b/src/core/constants/AVERAGE_BLOCKS_PER_10_MINS.ts
@@ -1,0 +1,11 @@
+import { AVERAGE_BLOCKS_PER_DAY } from './AVERAGE_BLOCKS_PER_DAY'
+import { Chain } from './chains'
+
+export const AVERAGE_BLOCKS_PER_10_MINUTES: Record<Chain, number> =
+  Object.entries(AVERAGE_BLOCKS_PER_DAY).reduce(
+    (acc, [chain, blocksPerDay]) => {
+      acc[chain as unknown as Chain] = (blocksPerDay * 10) / (24 * 60)
+      return acc
+    },
+    {} as Record<Chain, number>,
+  )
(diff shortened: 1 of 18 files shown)


src/core/constants/AVERAGE_BLOCKS_PER_10_MINS.ts
```diff
@@ -0,0 +1,11 @@
+import { AVERAGE_BLOCKS_PER_DAY } from './AVERAGE_BLOCKS_PER_DAY'
+import { Chain } from './chains'
+
+export const AVERAGE_BLOCKS_PER_10_MINUTES: Record<Chain, number> =
+  Object.entries(AVERAGE_BLOCKS_PER_DAY).reduce(
+    (acc, [chain, blocksPerDay]) => {
+      acc[chain as unknown as Chain] = (blocksPerDay * 10) / (24 * 60)
```
jpsains, line 7: since object keys are strings, indexing by "Chain" is a little annoying

src/core/utils/customJsonRpcProvider.ts
```diff
@@ -0,0 +1,33 @@
...
+   * @returns
+   */
+  async getStableBlockNumber(): Promise<number> {
+    const currentBlockNumber = await this.getBlockNumber()
+
+    const blockNumberTenMinsAgo =
+      currentBlockNumber - (AVERAGE_BLOCKS_PER_10_MINUTES[this.chainId] ?? 0) // default to 0 to avoid failures
+    return blockNumberTenMinsAgo
+  }
+}
```
jpsains, line 33: Im not 100% sure of the name wdyt? Do you prefer some of the below better?
"getTenMinuteOldBlockNumber"
"getSettledBlockNumber"

src/core/constants/AVERAGE_BLOCKS_PER_10_MINS.ts
bergarces: In other places of the code I have used `+chain as Chain` so I can compare it and pass the correct type down the line, but in this case it's going to convert them to a string for the key, so it doesn't matter.

src/core/utils/customJsonRpcProvider.ts
bergarces: The name sounds good to me. Better than the other suggestions.

There's also a TSDoc explaining what it does, so it should be clear.

bergarces (approved):
LGTM

PR #85, opened 2023-10-20
Add profits, deposits, withdrawals and tvl to Lido

bergarces:
- Fix patch version for jest-worker after critical vulnerability fix.
- Add protocol details to adapters.
- Fix prices for wstEth not taking into account block number passed.
- Implement stEth and wstEth adapters using `SimplePoolAdapter` (grants deposits, withdrawals, profits and TVL implementation).
- Remove metadata files from stEth and wstEth adapters, since it is easier to hardcode it.
- Change to `SimplePoolAdapter` to work with underlying ETH balances, not only ERC20.


src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -158,40 +159,39 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
...
-              ...underlyingTokenMetadata,
-              type: TokenType.Underlying,
-              totalSupplyRaw: underlyingTokenBalance,
-            }
-          }),
-        )
+        const [protocolTokenTotalSupply, underlyingTokenBalances] =
+          await Promise.all([
+            protocolTokenContact.totalSupply({ blockTag: blockNumber }),
+            getAddressesBalances({
```
bergarces, line 170: I cannot call `getBalances` directly, because that filters out 0 balances and changes the results of some aave snapshots in which the underlying balance is 0.

bergarces: In the previous version, I was assuming that underlying tokens where ERC20 contracts and getting their balances with `balanceOf`.

Because `stEth` underlying is `ETH`, that started throwing an error, which meant I had to do a special case for when then address was `ZERO_ADDRESS`. Instead of doing that, I remembered we had a nice method to do it, but that method stripped values that were 0.

However, there are multiple snapshots from aave-v2 and aave-v3, in which the total supply of the underlying is 0, and this was "breaking" the existing tests, as it was making the underlying disappear.

Thus why I need to call `getAddressesBalances` directly instead of `getBalances`.

For example:
```
{
          "address": "0xe91d55ab2240594855abd11b3faae801fd4c4687",
          "name": "Aave stable debt bearing USDT",
          "symbol": "stableDebtUSDT",
          "decimals": 6,
          "type": "protocol",
          "totalSupplyRaw": "3762995139075n",
          "tokens": [
            {
              "address": "0xdac17f958d2ee523a2206206994597c13d831ec7",
              "name": "Tether USD",
              "symbol": "USDT",
              "decimals": 6,
              "type": "underlying",
              "totalSupplyRaw": "0n",
              "totalSupply": "0.0",
              "iconUrl": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xdAC17F958D2ee523a2206206994597C13D831ec7/logo.png"
            }
          ],
          "totalSupply": "3762995.139075"
        }
```

jpsains (approved):

PR #86, opened 2023-10-23
feat: block number support

jpsains:


src/adapters/integration.test.ts
```diff
@@ -12,9 +13,9 @@ import { Protocol } from './protocols'
 import { testCases as stargateTestCases } from './stargate/tests/testCases'
 import { testCases as uniswapV3TestCases } from './uniswap-v3/tests/testCases'
 
-const TEST_TIMEOUT = 10000
+const TEST_TIMEOUT = 30000
```
bergarces, line 16: Ok for now as I've seen timeout errors too, but if we keep seeing errors in the pipeline, we should add some sort of retry mechanism for the tests that fail.

src/adapters/lido/tests/snapshots/ethereum.prices.json
```diff
@@ -59,8 +59,8 @@
               "symbol": "stETH",
               "decimals": 18,
               "type": "underlying",
-              "underlyingRateRaw": "1143460902465163371n",
-              "underlyingRate": "1.143460902465163371",
+              "underlyingRateRaw": "1143913215373500178n",
+              "underlyingRate": "1.143913215373500178",
```
bergarces, line 63: This will be fixed with my PR. Currently, the wsEth prices are not taking into consideration the block number.

src/core/utils/CustomMulticallJsonRpcProvider.ts
```diff
@@ -1,30 +1,31 @@
-import { TransactionRequest } from 'ethers'
+import { JsonRpcApiProviderOptions, TransactionRequest } from 'ethers'
 import { Chain } from '../constants/chains'
 import { CustomJsonRpcProvider } from './customJsonRpcProvider'
-import { logger } from './logger'
 import { MulticallQueue } from './multicall'
 
+export interface CustomTransactionRequest extends TransactionRequest {
+  blockTag?: number
```
bergarces, line 7: `blockTag?: BlockTag` is already part of `TransactionRequest`, and it already accepts `number` as a possible value.

This will only restrict it so that you can only pass a number, which can be annoying if you get a block number from a provider as a `bigint` or `string`.

Any reason why we need this interface?

src/core/utils/multicall.ts
```diff
@@ -79,47 +81,89 @@ export class MulticallQueue {
...
-      // reject all to be on safe side
-      callsToProcess.forEach(({ reject }) => {
-        reject('Multicall batch failed')
-      })
+      const resultLength = results.length
+
+      if (resultLength !== batchSize) {
+        // reject all to be on safe side
+        callsToProcess.forEach(({ reject }) => {
+          reject({ message: 'Multicall batch failed' })
```
bergarces, line 105: JS or TS don't enforce it (thus why the param for reject is `reason?: any`, but it is good practice to always throw/reject errors (which will indeed have `message: string` as per their interface,
```suggestion
          reject(new Error('Multicall batch failed'))
```

src/core/utils/multicall.ts
```diff
@@ -9,11 +10,14 @@ interface PendingCall {
     allowFailure: boolean
   }
   resolve: (value: string) => void
-  reject: (reason: string | undefined) => void
+  reject: (reason: { message: string }) => void
```
bergarces, line 13: I didn't realise before, but this is changing the signature of reject. It'd be better if it was `reason?: any`.

src/core/utils/multicall.ts
```diff
@@ -9,11 +10,14 @@ interface PendingCall {
...
 }
 
+type PendingCallsMap = Record<string | 'latest', PendingCall[]>
+
+const LATEST = 'latest'
```
bergarces, line 18:
```suggestion
const LATEST = 'latest'

type PendingCallsMap = Record<string | typeof LATEST, PendingCall[]>

```

PR #89, opened 2023-10-26
MMI-3993-enforce-unique-product-id

bergarces:


src/scripts/newAdapterCommand.ts
```diff
@@ -117,14 +121,52 @@ export function newAdapterCommand(program: Command) {
               isKebabCase(input) || 'Value must be kebab-case',
             when: !inputProtocolId,
           },
+          {
+            type: 'checkbox',
+            name: 'chainKeys',
+            message: 'What chains will the adapter be valid for?',
+            choices: Object.keys(Chain),
+            default: chainKeys.length ? chainKeys : ['Ethereum'],
+          },
```
bergarces, line 130: Needs to be moved before `productId` is defined, as it is possible to have two different adapters with the same `productId` as long as they don't overlap the chains for which they are being used.

jpsains (approved):

PR #91, opened 2023-10-31
feat: curve

jpsains:


src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -38,54 +51,83 @@ export class CurvePoolAdapter
...
+
+    const prices = await this.getUnderlyingTokenConversionRate(
+      protocolToken,
+      blockNumber,
+    )
+
+    return prices.map((underlyingTokenPriceObject) => {
+      const underlyingRateRawBigInt = BigInt(
+        underlyingTokenPriceObject.underlyingRateRaw,
+      )
```
bergarces, line 115: This is already a `bigint`, it doesn't need conversion.

src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -38,54 +51,83 @@ export class CurvePoolAdapter
...
+      const balanceRawBigInt = BigInt(protocolTokenBalance.balanceRaw)
```
bergarces, line 116: Same as before, already a `bigint`.

src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -38,54 +51,83 @@ export class CurvePoolAdapter
...
   protected async getUnderlyingTokenBalances(
-    _protocolTokenBalance: TokenBalance,
-    _blockNumber?: number,
+    protocolTokenBalance: TokenBalance,
+    blockNumber: number,
   ): Promise<Underlying[]> {
-    throw new NotImplementedError()
+    const { protocolToken } = await this.fetchPoolMetadata(
+      protocolTokenBalance.address,
+    )
```
bergarces, line 105: nit: You can use `this.fetchProtocolTokenMetadata`, which is already implemented.

src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -133,15 +171,46 @@ export class CurvePoolAdapter
...
+
+    const lpTokenContract = Erc20__factory.connect(
+      protocolToken.address,
+      this.provider,
+    )
+
+    const supply = await lpTokenContract.totalSupply({ blockTag: blockNumber })
+
+    return balances.slice(0, underlyingTokens.length).map((balance, index) => {
```
bergarces, line 198: I'm failing to see the logic behind slicing the array here.

src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -133,15 +171,46 @@ export class CurvePoolAdapter
...
+
+      const price = +formattedBalance / +formattedSupply
+
+      return {
+        type: TokenType.Underlying,
+        underlyingRateRaw: BigInt(
+          Math.round(price * 10 ** underlyingToken.decimals),
+        ),
+        ...underlyingToken,
+      }
```
bergarces, line 212: Ok...

So the balance and the supply are being converted first to "numbers" (really to strings, then to numbers, but that makes no difference) and then multiplied again by the decimals and converted to `bigint`. Very dodgy.

To calculate `formatBalance`, what's essentially doing is `balance / 10 ** underlyingToken.decimals`, and then at the end we are multiplying by `10 ** underlyingToken.decimals`, which I think means we could skip the first division.

What I suggest we do instead:

```suggestion
      const formattedBalance = formatUnits(balance, underlyingToken.decimals)
      const formattedSupply = formatUnits(supply, protocolToken.decimals)

      const underlyingRateRaw = balance / (supply / 10 ** protocolToken.decimals)

      return {
        type: TokenType.Underlying,
        underlyingRateRaw,
        ...underlyingToken,
      }
```

src/adapters/integration.test.ts
```diff
@@ -27,6 +27,7 @@ function runAllTests() {
   runProtocolTests(Protocol.UniswapV3, uniswapV3TestCases)
   runProtocolTests(Protocol.Lido, lidoTestCases)
   runProtocolTests(Protocol.Curve, curveTestCases)
+  runProtocolTests(Protocol.Curve, curveTestCases)
```
bergarces, line 30: This is running curve tests twice.

src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -187,4 +248,58 @@ export class CurvePoolAdapter
...
+    )
+    const poolAddress = await metaRegistryContract.pool_list(i)
+    const lpToken = await metaRegistryContract['get_lp_token(address)'](
+      poolAddress,
+    )
+    const lpTokenContract = Erc20__factory.connect(lpToken, this.provider)
+
+    const underlyingCoins = (
+      await metaRegistryContract['get_underlying_coins(address)'](poolAddress)
+    ).filter((address) => address !== ZERO_ADDRESS)
```
bergarces, line 272: I've noticed in their metadata that they use 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee for ETH. Does that need checking here?

src/core/metadata/token-metadata-ethereum.json
```diff
@@ -74298,8 +74305,8 @@
     "decimals": 18,
     "iconUrl": "https://raw.githubusercontent.com/MetaMask/contract-metadata/master/images/Genieswap.svg"
   },
-  "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee": {
-    "address": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
+  "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE": {
```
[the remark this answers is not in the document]
jpsains: Curve seem to use this address not seen other protocols used it so far.

src/adapters/curve/products/pool/curvePoolAdapter.ts
jpsains: I believe they use Zero_address to represent empty slots in an array because vyper language does not have dynamic arrays

src/adapters/curve/products/pool/curvePoolAdapter.ts
jpsains: ok probably: no dynamic array limitation of vyper so they set the response of the balances to an array which does not match with the amount of underlying tokens, nonetheless Ive switched the implementation.

jpsains: Ill iterate over the underlying token array instead

PR #97, opened 2023-11-17
GMX Adapter

bergarces:


src/adapters/gmx/products/glp/gmxGlpAdapter.ts
```diff
@@ -373,4 +359,76 @@ export class GMXGlpAdapter implements IProtocolAdapter, IMetadataBuilder {
...
+    blockNumber?: number | undefined
+  }): Promise<Underlying[]> {
+    throw new NotImplementedError()
+  }
+  protected getUnderlyingTokenConversionRate(
+    _protocolTokenMetadata: Erc20Metadata,
+    _blockNumber?: number | undefined,
+  ): Promise<UnderlyingTokenRate[]> {
+    throw new NotImplementedError()
+  }
```
bergarces, line 433: We are going to have a problem with `SimplePoolAdapter`.

In this case, I want to extend because I want to use its `getMovements`.
On the other hand, I don't want to use it for `getPositions` or `getPrices`, but the class forces me to implement this methods.

We might have to separate things a bit so that we can cherrypick. We can't have multiple inheritance, but we can have classes that partially implement some methods.

src/adapters/gmx/products/glp/gmxGlpAdapter.ts
```diff
@@ -373,4 +359,76 @@ export class GMXGlpAdapter implements IProtocolAdapter, IMetadataBuilder {
...
+        protocolTokenAddress,
+      ),
+      underlyingTokens: await this.fetchUnderlyingTokensMetadata(
+        protocolTokenAddress,
+      ),
+      filter: {
+        smartContractAddress: protocolTokenAddress,
+        fromBlock,
+        toBlock,
+        from: ZERO_ADDRESS,
```
bergarces, line 380: In this case, it needs to be from `ZERO_ADDRESS`, not `undefined`.

Otherwise, given how the GLP token is deposited, it'll count both as a deposit and a withdrawal.

https://arbiscan.io/tx/0x47fbf2b3e9fc2c56d266469cb39f25b5e1404d96ad2d2a54d9f3838cb834e095

src/core/adapters/SimplePoolAdapter.ts
```diff
@@ -424,7 +424,7 @@ export abstract class SimplePoolAdapter implements IProtocolAdapter {
               const movementValueRaw =
                 (protocolTokenMovementValueRaw *
                   currentTokenPrice.underlyingRateRaw) /
-                BigInt(10 ** currentTokenPrice.decimals)
+                BigInt(10 ** protocolToken.decimals)
```
bergarces, line 427: This was wrong before, but we have not noticed due to decimals being the same.

The numbers that we multiply have their own decimals, and we need to divide by both for the final result. Since the movement value if for the underlying, we need to divide by the protocol token. The final step outside the adapter divides by the underlying.

src/adapters/stargate/tests/snapshots/ethereum.profits.json
```diff
@@ -111,19 +111,19 @@
...
               "symbol": "DAI",
               "decimals": 18,
-              "profitRaw": "-18179952999981820051n",
+              "profitRaw": "-3229072000000n",
               "type": "underlying",
               "calculationData": {
-                "withdrawalsRaw": "18179949n",
-                "withdrawals": "0.000000000018179949",
+                "withdrawalsRaw": "18179949770928000000n",
+                "withdrawals": "18.179949770928",
```
bergarces, line 118: This is the correct value according to what occurred in this tx (https://etherscan.io/tx/0x110172990ef6a644b958a07051d2bc8b52e17db3c39f2a1cdbf7d8686c598e52)

The following command isolates all the withdrawals for this protocol token within the blocks specified.

```
npm run withdrawals 0xB0D502E938ed5f4df2E681fE6E419ff29631d62b 18156819 18163965 0x0faf1d2d3ced330824de3b8200fc8dc6e397850d pool stargate 1 12345
```

The wrong value was due to `SimplePoolAdapter` movements calculation dividing by the wrong number of decimals.

src/adapters/gmx/products/glp/gmxGlpAdapter.ts
jpsains: ok a solution inst jumping out to me atm, perhaps we want to separate the profit methods e.g. getProfits and getMovements into a separate class.....not sure

src/adapters/gmx/tests/snapshots/arbitrum.positions.json
```diff
@@ -0,0 +1,129 @@
...
+          "balanceRaw": "1788213725943696140789934n",
+          "tokens": [
+            {
+              "address": "0x2f2a2543b76a4166549f7aab2e75bef0aefc5b0f",
+              "name": "Wrapped BTC",
+              "symbol": "WBTC",
+              "decimals": 8,
+              "type": "underlying",
+              "balanceRaw": "1661250551n",
+              "balance": "16.61250551",
```
jpsains, line 30: the numbers seem to be significantly more than what debank is showing 13 vs to 16 btc (your results) https://debank.com/profile/0xbdfa4f4492dd7b7cf211209c4791af8d52bf5c50

jpsains: Thoughts, I guess yours are from 4 days ago

jpsains: image

src/adapters/gmx/tests/snapshots/arbitrum.profits.json
```diff
@@ -0,0 +1,229 @@
...
+              "iconUrl": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/arbitrum/assets/0x2f2a2543B76A4166549F7aaB2e75Bef0aefC5B0f/logo.png"
+            }
+          ]
+        }
+      ],
+      "fromBlock": 148803082,
+      "toBlock": 151045028
+    }
+  ]
+}
```
jpsains, line 229: results look reasonable

src/adapters/stargate/tests/snapshots/ethereum.profits.json
jpsains: thanks for checking

jpsains (approved):

jpsains:
LGTM next task would be to get this value and previous "withdrawals" related to "claimedRewards"
image

src/adapters/gmx/tests/snapshots/arbitrum.positions.json
bergarces: It's due to it being a few days old. Running the command for latest block returns similar results.

image

PR #98, opened 2023-11-17
feat: reward

jpsains:


src/core/utils/mergeClaimableWithProtocol.ts
```diff
@@ -0,0 +1,80 @@
+import { DefiPositionResponse } from "../../defiProvider"
+import { Underlying, ProtocolPosition } from "../../types/adapter"
+import { DisplayPosition } from "../../types/response"
+
+
+export function mergeClaimableWithProtocol(positions: DefiPositionResponse[]) {
```
jpsains, line 6: Happy to walk you through this, refer to uni tests for more context

src/core/utils/mergeClaimableWithProtocol.ts
```diff
@@ -0,0 +1,80 @@
...
+      const mergedTokens = position.tokens.map((lpPosition) => {
+        const rewards = rewardMap[lpPosition.address]
+        const mergedTokens = rewards
+          ? [...rewards, ...(lpPosition.tokens || [])]
+          : lpPosition.tokens
+        return { ...lpPosition, tokens: mergedTokens }
+      })
+      mergedResponse.push({
+        ...position,
+        tokens: (mergedTokens as DisplayPosition<ProtocolPosition>[]) || [],
```
jpsains, line 41: Feel free to propose better solution here rather than casting.

src/adapters/curve/products/reward/curveRewardAdapter.ts
```diff
@@ -0,0 +1,130 @@
...
+      const stakingContract = StakingContract__factory.connect(
+        address,
+        this.provider,
+      )
+
+      const balanceRaw = await stakingContract.claimable_tokens.staticCall(
+        userAddress,
+        { blockTag: blockNumber },
+      ).catch(e => 0n)
+
```
jpsains, line 69: Latest curve pools (which have low liquidity) have a different reward interface. So this requests fails for those pools. 

Since with this MR we have feature parity with our old adapters I think we should manage new curve-rewards under a different adapter.

jpsains: Upon further inspection, they do have this method, it fails if you run the tests at a block number before the contract was deployed

jpsains: catching and returning 0 I think is not a bad solution...

src/core/utils/mergeClaimableWithProtocol.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export function mergeClaimableWithProtocol(positions: DefiPositionResponse[]) {
+  const rewardMap: Record<string, DisplayPosition<Underlying>[]> = {}
+  const nonRewardPositions: DefiPositionResponse[] = []
+  const mergedResponse: DefiPositionResponse[] = []
+
+  // Build the reward map and separate non-reward positions
+  positions.forEach((position) => {
+    if (position.success) {
+      if (position.positionType === 'reward') {
```
bergarces, line 14:
```suggestion
      if (position.positionType === PositionType.reward) {
```

src/core/utils/mergeClaimableWithProtocol.ts
```diff
@@ -0,0 +1,80 @@
...
+        position.tokens.forEach(({ address: lpTokenAddress, tokens, type }) => {
+          if (type === 'claimable') {
```
bergarces, line 16:
```suggestion
          if (type === TokenType.Reward) {
```

bergarces (approved):
I can't fault it.

I've been desperately thinking on how to avoid having to create a new adapter for the rewards, as they share so much with the protocol adapter and the position results are going to be merged anyway.

Even though I think it's relatively easy to add the rewards along with the positions without creating a new adapter, the same cannot be said for withdrawals and profits, so I guess we'll have to stick with this for now.

PR #100, opened 2023-11-20
Mmi-4190-template-updates

jpsains:


bergarces (approved):
Nice removing code. :D

PR #101, opened 2023-11-21
recursive calculation of underlying protocol tokens

bergarces:
- Added method to AdaptersController to create a list of protocol tokens per chain
- Added decorator that recursively computes underlying positions
- Fixed wrong metadata symbol for lido protocol tokens.

### DIFF
--- a/src/adapters/lido/products/st-eth/lidoStEthAdapter.ts
+++ b/src/adapters/lido/products/st-eth/lidoStEthAdapter.ts
@@ -48,7 +48,7 @@ export class LidoStEthAdapter extends SimplePoolAdapter {
     return {
       address: '0xae7ab96520de3a18e5e111b5eaab095312d7fe84',
       name: 'Liquid staked Ether 2.0',
-      symbol: 'stETH',
+      symbol: 'STETH',
       decimals: 18,
     }
   }
--- a/src/adapters/lido/products/st-eth/lidoStEthAdapter.ts
+++ b/src/adapters/lido/products/st-eth/lidoStEthAdapter.ts
@@ -48,7 +48,7 @@ export class LidoStEthAdapter extends SimplePoolAdapter {
     return {
       address: '0xae7ab96520de3a18e5e111b5eaab095312d7fe84',
       name: 'Liquid staked Ether 2.0',
-      symbol: 'STETH',
+      symbol: 'stETH',
       decimals: 18,
     }
   }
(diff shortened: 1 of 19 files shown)


src/adapters/lido/products/st-eth/lidoStEthAdapter.ts
```diff
@@ -45,22 +46,11 @@ export class LidoStEthAdapter extends SimplePoolAdapter {
...
 
   protected async fetchProtocolTokenMetadata(): Promise<Erc20Metadata> {
-    return {
-      address: '0xae7ab96520de3a18e5e111b5eaab095312d7fe84',
-      name: 'Liquid staked Ether 2.0',
-      symbol: 'STETH',
-      decimals: 18,
-    }
+    const stEthAddress = '0xae7ab96520de3a18e5e111b5eaab095312d7fe84'
+    return await getTokenMetadata(stEthAddress, this.chainId, this.provider)
```
bergarces, line 50: Using `getTokenMetadata` to avoid mismatches with our metadata list.

src/core/adaptersController.ts
```diff
@@ -67,6 +70,52 @@ export class AdaptersController {
...
+  ): Promise<IProtocolAdapter | undefined> {
+    // Only run this first time
+    if (this.protocolTokens.size === 0) {
+      for (const [chainId, chainAdapters] of this.adapters) {
+        this.protocolTokens.set(chainId, new Map())
+        const chainAdaptersMap = this.protocolTokens.get(chainId)!
+
+        for (const [_protocolId, protocolAdapters] of chainAdapters) {
+          for (const [_productId, adapter] of protocolAdapters) {
+            // TODO Uncomment when new type is merged
```
bergarces, line 85: This code cannot be added until the newreward type is merged.

src/types/IProtocolAdapter.ts
```diff
@@ -35,6 +36,8 @@ export interface IProtocolAdapter {
    */
   productId: string
 
+  adaptersController: AdaptersController
```
bergarces, line 39: As discussed, I think it would be beneficial to stop adding things here at public level and add instead a single toolbox class that exposes what we need:
- Methods from adapter controller
- ChainId
- Provider
- Anything new we want to add

If we do that, then we won't have to change every adapter that implements this interface directly every time we add a utility to that class (unless that adapter needs to use it).

jpsains (approved):
LGTM I uncommented the if(reward) continue

PR #103, opened 2023-11-23
add adapter for swell sweth

dannydalton-swell:
This PR adds an adapter for Swell Ethereum (swETH). It is an LST similar to Lido's stETH. Used the Lido adapter as an example with some slight changes to implement this one.


jpsains:
> This PR adds an adapter for Swell Ethereum (swETH). It is an LST similar to Lido's stETH. Used the Lido adapter as an example with some slight changes to implement this one.

bergarces (approved):

dannydalton-swell:
thanks guys!

PR #105, opened 2023-11-24
feat: prices

jpsains:


src/adapters/prices/products/prices/prices.json
jpsains, line 1: Results from my script.

See hasPrice and noPrice inside file below.

Has about 300 token prices.

Note some prices have not been updated for a while but they might be sh*t coins

jpsains:
https://github.com/consensys-vertical-apps/mmi-defi-adapters/blob/db95e49144224e65e5576f3ac2b839c42f2582f2/src/adapters/uniswap-v3/tests/snapshots/ethereum.positions.json#L55

👋 @bergarces 

Does the contract address need to be lower case for the token to resolve? It didn't resolve here for some reason

bergarces:
> https://github.com/consensys-vertical-apps/mmi-defi-adapters/blob/db95e49144224e65e5576f3ac2b839c42f2582f2/src/adapters/uniswap-v3/tests/snapshots/ethereum.positions.json#L55
> 
> 👋 @bergarces
> 
> Does the contract address need to be lower case for the token to resolve? It didn't resolve here for some reason

It doesn't need to be.

PR #112, opened 2023-12-07
feat: profit

jpsains:
MR includes the following 

- Change to the MovementByBlock interface to make it similar to getPositions response and new design was easier to "resolve" recursivly 
- New decarator which resolves movements, therefore applies fiat(USD) value to movements.
- Update getProfits to aggrigate fiat(USD) values instead of token balances

### DIFF
--- a/src/adapters/convex/tests/testCases.ts
+++ b/src/adapters/convex/tests/testCases.ts
@@ -39,7 +39,7 @@ export const testCases: TestCase[] = [
       protocolTokenAddress: '0x30d9410ed1d5da1f6c8391af5338c93ab8d4035c',
       productId: 'pool',
     },
-  }, //  npm run deposits 0xdf286De6d3de10A6aD6452d0BA94Af7AD7B68F9B 14443070 14443071 0x30d9410ed1d5da1f6c8391af5338c93ab8d4035c pool curve 1
+  },
   {
     chainId: Chain.Ethereum,
     method: 'withdrawals',
--- a/src/adapters/curve/products/pool/curvePoolAdapter.ts
+++ b/src/adapters/curve/products/pool/curvePoolAdapter.ts
@@ -43,8 +43,7 @@ type CurvePoolAdapterMetadata = Record<
 
 export class CurvePoolAdapter
   extends SimplePoolAdapter
-  implements IMetadataBuilder
-{
+  implements IMetadataBuilder {
   productId = 'pool'
 
   getProtocolDetails(): ProtocolDetails {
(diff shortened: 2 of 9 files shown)


src/adapters/convex/tests/snapshots/ethereum.deposits.json
bergarces, line 1: Should we just get rid of this test case?

src/adapters/convex/tests/testCases.ts
```diff
@@ -39,7 +39,7 @@ export const testCases: TestCase[] = [
       protocolTokenAddress: '0x30d9410ed1d5da1f6c8391af5338c93ab8d4035c',
       productId: 'pool',
     },
-  },
+  }, //  npm run deposits 0xdf286De6d3de10A6aD6452d0BA94Af7AD7B68F9B 14443070 14443071 0x30d9410ed1d5da1f6c8391af5338c93ab8d4035c pool curve 1
```
bergarces, line 42: We can delete this comment.

src/adapters/curve/products/pool/curvePoolAdapter.ts
```diff
@@ -161,6 +167,10 @@ export class CurvePoolAdapter
       this.provider,
     )
 
+    if (this.chainId == 1 && blockNumber && blockNumber < 15732062) {
```
bergarces, line 170:
```suggestion
    if (this.chainId == Chain.Ethereum && blockNumber && blockNumber < 15732062) {
```
Best to use the constant.

src/adapters/curve/tests/snapshots/ethereum.profits.profits-2.json
```diff
@@ -24,40 +24,13 @@
...
-                "startPositionValueRaw": "0n",
-                "startPositionValue": "0.0",
-                "endPositionValueRaw": "6357359611959254406n",
-                "endPositionValue": "6.357359611959254406"
-              },
-              "profit": "0.0",
-              "iconUrl": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490/logo.png"
-            }
-          ]
+      "success": false,
```
bergarces, line 27: This is a weird one.

It has deposits and an end position, but it didn't resolve the underlyings (DAI, USDC and USDT).

Maybe a job for another ticket.

bergarces (dismissed):

src/adapters/convex/tests/snapshots/ethereum.deposits.json
jpsains: I think we could.

I might just keep it until the repo settles down 

Yesterday it came in handy because the result changed because I had lost blockNumber iini the resolver

src/adapters/curve/tests/snapshots/ethereum.profits.profits-2.json
jpsains: its fixed now, I believe one of the adapters was missing the decorator

PR #113, opened 2023-12-07
prevent provider errors from stopping process

bergarces:


src/core/utils/chainProviders.ts
```diff
@@ -8,6 +8,23 @@ import { CustomMulticallJsonRpcProvider } from './CustomMulticallJsonRpcProvider
...
+const ChainMulticallSettings: Record<
+  Chain,
+  { flushTimeoutMs: number; maxBatchSize: number }
+> = Object.values(Chain).reduce(
+  (accumulator, chainId) => {
+    return {
+      [chainId]: {
+        flushTimeoutMs: 0.1,
+        // Allow a bigger batch size for mainnet
+        maxBatchSize: Chain.Ethereum ? 200 : 100,
```
bergarces, line 20: Making it configurable via env is possible, but we would need to pass some sort of json string with the values per chain.

In the meantime, I think that reducing the batch size for non-mainnet chains is a good policy.

src/core/utils/multicall.ts
```diff
@@ -89,28 +94,67 @@ export class MulticallQueue {
...
+          },
+          'Multicall error when sending a batch',
+        )
+
+        return
+      }
 
       const resultLength = results.length
 
       if (resultLength !== batchSize) {
```
bergarces, line 134: We shouldn't remove this, as I actually managed to trigger it multiple times by increasing the batch size to several thousands.

jpsains: Oh wow. Thats odd, i didnt think it would be possible

jpsains (approved):

PR #133, opened 2023-12-20
Mmi 4360 add protocol token address filter

bergarces:
- Adds possibility to filter positions or profits by protocol token addresses.
- Improves stress tests to include optional parameters (chain and protocol filters, use specific address, number of iterations or strategy


src/adapters/convex/products/cvxcrv-wrapper/convexCvxcrvWrapperAdapter.ts
```diff
@@ -61,10 +61,19 @@ export class ConvexCvxcrvWrapperAdapter
   async getPositions({
     userAddress,
     blockNumber,
+    protocolTokenAddresses,
   }: GetPositionsInput): Promise<ProtocolPosition[]> {
-    const [protocolTokens] = await this.getProtocolTokens()
+    const [protocolToken] = await this.getProtocolTokens()
```
jpsains, line 66: i think your idea about passing the protocolTokenAddresses to this.getProtocolTokens() was a nice idea. Anyway lets do it later

PR #138, opened 2024-01-08
new adapter test

bergarces:
Split as two jobs so that tests can start running without waiting for linting and building new adapters.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
-      
+
       - name: Install dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test.yaml
bergarces, line 1: Creates an adapter with every available template.

If the template implements `IMetadataBuilder` like `SimplePoolAdapter`, it also needs to build en empty metadata file or the tests fail.

src/scripts/newAdapterCommand.ts
bergarces, line 1: Adds a `-y` option to skip questions if the values have already been provided (similar to `npm init -y`).

This is needed in order to test the command without having to provide additional input.

jpsains (approved):

PR #139, opened 2024-01-09
MMI-4401-fix-build-metadata

bergarces:


src/scripts/buildMetadata.ts
```diff
@@ -63,26 +63,28 @@ export function buildMetadata(
...
-                metadata,
-              })
+            await writeMetadataToFile({
+              ...fileDetails,
+              metadata,
+            })
 
-              addStaticImport(fileDetails)
-            }
+            await addStaticImport(fileDetails)
```
bergarces, line 87: This was the first issue that was causing `AdaptersMetadata.ts` file to become corrupt, since we were not awaiting, sometimes two process would write at the same time.

jpsains (dismissed):

PR #147, opened 2024-01-22
MMI-4477-investigate-aave-profit-numbers-on-dashboard

jpsains:


src/core/getProfits.ts
```diff
@@ -35,33 +44,47 @@ export async function getProfits({
...
-        .then(aggregateFiatBalances),
+        .then((result) => {
+          rawStartPositionValues = result
+          return aggregateFiatBalances(result)
+        }),
     ])
   } else {
     // Call toBlock first and filter fromBlock
-    endPositionValues = await adapter
+    ;(endPositionValues = await adapter
```
bergarces, line 64: Is this intended? I mean the `;(` part.

In this branch, we want to first run the adapter with the end positions, and then pass those positions as a filter.

src/core/getProfits.ts
```diff
@@ -103,7 +128,9 @@ export async function getProfits({
       const profit =
         endPositionValue + withdrawal - deposit - startPositionValue
 
-      if (adapter.getProtocolDetails().positionType == PositionType.Borrow) {
+      const isBorrow =
+        adapter.getProtocolDetails().positionType == PositionType.Borrow
+      if (isBorrow) {
         profit * -1
```
bergarces, line 134: Weird that we didn't caught this before but... this expression is not being assigned anywhere.

src/core/getProfits.ts
```diff
@@ -120,9 +147,29 @@ export async function getProfits({
         calculationData: {
           withdrawals: withdrawal,
           deposits: deposit,
-          startPositionValue: startPositionValue,
-          endPositionValue: endPositionValue,
+          startPositionValue: isBorrow
+            ? startPositionValue * -1
+            : startPositionValue,
+          endPositionValue: isBorrow ? endPositionValue * -1 : endPositionValue,
```
bergarces, line 153: In order to avoid repeating this logic branching, we can create a `profitModifier` and just set it up once.

```
const profitModifier =
        adapter.getProtocolDetails().positionType === PositionType.Borrow
          ? -1
          : 1
 ```
 
 Then we just always multiply by that value.

bergarces (approved):

PR #148, opened 2024-01-22
retry handler and refactor

bergarces:
All provider related files have been moved to their own folder outside `utils`. Some of them have been renamed to match what they export.

Two new environment variables are added with defaults:
- One for setting the amount of retries (defaults to 1)
- One for setting the time for the call to timeout in ms (defaults to 20s)

The retry/timeout handler is only applied to `CustomJsonRpcProvider` call method.

When using `CustomMulticallJsonRpcProvider`, it'll either:
- Use the `MulticallQueue`, which has been initialised with a contract using `CustomJsonRpcProvider`
- Use `super.call`, which is the implementation of `CustomJsonRpcProvider`


src/config.ts
```diff
@@ -67,6 +67,14 @@ const ConfigSchema = z
...
       ),
+    rpcCallTimeoutInMs: z
+      .number()
+      .default(
+        Number(process.env.DEFI_ADAPTERS_RPC_CALL_TIMEOUT_IN_MS) || 20000,
+      ),
+    rpcCallRetries: z
+      .number()
+      .default(Number(process.env.DEFI_ADAPTERS_RPC_CALL_RETRIES) || 1),
   })
```
jpsains, line 78: worth updating the example.env file?

src/core/provider/CustomJsonRpcProvider.ts
```diff
@@ -29,4 +43,8 @@ export class CustomJsonRpcProvider extends JsonRpcProvider {
       currentBlockNumber - (AVERAGE_BLOCKS_PER_10_MINUTES[this.chainId] ?? 0) // default to 0 to avoid failures
     return blockNumberTenMinsAgo
   }
+
+  async call(transaction: TransactionRequest): Promise<string> {
+    return this.retryHandler(() => super.call(transaction))
+  }
 }
```
jpsains, line 50: can you add the retry handler on getLogs aswel please?

jpsains: i wouldn't wait 20 second for logs though the request should be resolved in a few seconds

bergarces: Added another `retryHandler` for `getLogs`.

jpsains (approved):

PR #150, opened 2024-01-23
feat: address checksum

jpsains:


src/adapters/chimp-exchange/products/pool/chimpExchangePoolAdapter.ts
```diff
@@ -175,9 +176,7 @@ export class ChimpExchangePoolAdapter
         const tokensData = await filterMapAsync(
           event.args.tokens,
           async (token, index) => {
-            if (
-              token.toLowerCase() === protocolTokenAddress.toLocaleLowerCase()
-            ) {
+            if (getAddress(token) === getAddress(protocolTokenAddress)) {
```
bergarces, line 179: I think `protocolTokenAddress` here should already be checksumed, as it's input from defiProvider.

bergarces (approved):
I'll make sure to add to the PR check list that addresses stored on metadata, hardcoded or coming from SCs (unless they are sure they are checksumed) should be passed through `getAddress` method.

PR #158, opened 2024-02-02
fork and simulate

bergarces:
Added ability to simulate a write adapter transaction result.

- Add `--simulate`
- Include `--user-address` or `-u` followed by the address
- Include `--token-address` or `-t`, which is the address of the protocol token
- Include `--asset` or `-a`, which is the address of the underlying token. This is optional and it sends a transaction to approve that token.

Example:
```
npm run params -- supply aave-v3 a-token 1 0x6B175474E89094C44Da98b954EedeAC495271d0F,100000000000000000000,0xfECA27229A3801b34591582B66C46c82a9bC9994,0 --simulate --user-address 0xfECA27229A3801b34591582B66C46c82a9bC9994 --token-address 0x018008bfb33d285247A21d44E50697654f754e63 --asset 0x6B175474E89094C44Da98b954EedeAC495271d0F
```

Also includes commands to simulate transactions that have already ocurred using the txHash.

Example:
```
npm run simulate -- 0x669fc9c7a762e0235b23967d5437a91623b7193a04b173239d80ac988eecc372 1 0x98C23E9d8f34FEFb1B7BD6a91B7FF122F4e16F5c aave-v3 a-token -b 19133680
```


src/core/decorators/cacheToFile.ts
```diff
@@ -64,16 +64,6 @@ export function CacheToFile({ fileKey }: { fileKey: string }) {
         throw new Error('Metadata file not found')
       }
 
-      logger.debug(
```
bergarces, line 67: This made the logs impossible to read, and the more files we add the worse it'll get. If file could not be loaded it would error anyway.

src/core/utils/getTokenMetadata.ts
```diff
@@ -24,10 +24,6 @@ export async function getTokenMetadata(
   if (fileMetadata) {
     const fileTokenMetadata = fileMetadata[tokenAddress.toLowerCase()]
     if (fileTokenMetadata) {
-      logger.debug(
-        { tokenAddress, chainId },
-        'Token metadata found on cached file',
-      )
```
bergarces, line 30: Spam logs

PR #166, opened 2024-02-20
feat: missing profits

jpsains:
- change to the way we handle missing prices on profits.
- Before we threw when one price was missing.
- Now we return positions and indicate if there is a price missing in the response.
- Also, handle when price is 0, before was returned as undefined instead of 0
- Also, bug fix which was stopping positions with reward tokens from returning.

### DIFF
--- a/src/adapters/convex/products/extra-reward/convexExtraRewardAdapter.ts
+++ b/src/adapters/convex/products/extra-reward/convexExtraRewardAdapter.ts
@@ -328,7 +328,15 @@ export class ConvexExtraRewardAdapter
     const poolMetadata = (await this.buildMetadata())[protocolTokenAddress]
 
     if (!poolMetadata) {
-      logger.error({ protocolTokenAddress }, 'Protocol token pool not found')
+      logger.warn(
+        {
+          protocolTokenAddress,
+          protocol: this.protocolId,
+          chainId: this.chainId,
+          product: this.productId,
+        },
+        'Protocol token pool not found',
+      )
       throw new Error('Protocol token pool not found')
     }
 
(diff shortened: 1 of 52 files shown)


src/adapters/convex/products/extra-reward/convexExtraRewardAdapter.ts
```diff
@@ -328,7 +328,15 @@ export class ConvexExtraRewardAdapter
     const poolMetadata = (await this.buildMetadata())[protocolTokenAddress]
 
     if (!poolMetadata) {
-      logger.error({ protocolTokenAddress }, 'Protocol token pool not found')
+      logger.warn(
```
bergarces, line 331: `logger.error`

src/core/decorators/addClaimableRewards.ts
```diff
@@ -27,18 +27,25 @@ export function AddClaimableRewards({
...
                     ...(protocolToken.tokens ?? []),
                     ...reward.tokens,
                   ]
                 }
+              } catch (error) {
+                if (
+                  (error as Error).message === 'Protocol token pool not found'
+                ) {
+                  return
+                }
```
bergarces, line 47:
```suggestion
                if (error instaceOf Error && error.message === 'Protocol token pool not found') {
                  return
                }
```

If the error is not an instance of error for whatever reason and is just a string (it can happen), `error.message` would crash.

My worry is that error is not really standarised, I believe we only use it for the `SimplePoolAdapter`, not even for the `DefaultAdapter`, but for now this should do the trick.

src/core/utils/aggregateFiatBalances.ts
bergarces, line 1: mmmmm

I think we merge this to fix the bugs, but we might want to revisit the code eventual, a lot of optional fields and things.

bergarces (approved):

src/adapters/convex/products/extra-reward/convexExtraRewardAdapter.ts
jpsains: The issue here is not every protocol token will have rewards available.

jpsains: So we were getting error thrown

PR #168, opened 2024-02-23
Snapshot test filter improvements

bergarces:


src/adapters/prices-v2/tests/snapshots/ethereum.prices.json
```diff
@@ -1,5 +1,5 @@
 {
-  "blockNumber": 19283451,
+  "blockNumber": 19311615,
```
bergarces, line 2: Quite cool, this has been picked up by the new snapshot builder and now it has added the block number to the testcase as well. So it's working for both protocolTokenAddresses and blockNumber.

jpsains (approved):

PR #175, opened 2024-02-29
Mmi 4532 add curve and convex support on abritrum

jpsains:


src/core/utils/getTokenMetadata.ts
```diff
@@ -20,6 +21,23 @@ export async function getTokenMetadata(
   chainId: Chain,
   provider: CustomJsonRpcProvider,
 ): Promise<Erc20Metadata> {
+  if (tokenAddress === '0x6b8734ad31D42F5c05A86594314837C416ADA984') {
```
bergarces, line 24: Where does this address come from? Should we put a comment or something for future readers?

PR #182, opened 2024-03-07
MMI-4724-support-for-compound-v-2-borrows

bergarces:


src/core/utils/filters.ts
bergarces, line 1: I broaden the type narrower because it wouldn't let me pass readonly arrays that come from ethers log parsing functions.

This way it accepts both readonly and regular arrays.

src/adapters/compound-v2/products/borrow-market/compoundV2BorrowMarketAdapter.ts
bergarces, line 1: Implements directly `IProtocolAdapter`.

src/adapters/compound-v2/products/borrow-market/compoundV2BorrowMarketAdapter.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    const pools = await comptrollerContract.getAssetsIn(userAddress)
+
+    return await filterMapAsync(pools, async (poolContractAddress) => {
+      if (
+        protocolTokenAddresses &&
+        !protocolTokenAddresses.includes(poolContractAddress)
+      ) {
+        return undefined
+      }
```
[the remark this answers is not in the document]
bergarces: There's not much choice in this case.

It's a filter, so you either leave it as undefined and don't use it, or if it's instantiated as an array you filter using it.

If we set it to default to `[]`, instead of checking for it not to be undefined, we would need to check that the length is not `0`, which is more cumbersome.

src/adapters/compound-v2/products/borrow-market/compoundV2BorrowMarketAdapter.ts
```diff
@@ -0,0 +1,355 @@
...
+    throw new NotImplementedError()
+  }
+
+  getWithdrawals(_input: GetEventsInput): Promise<MovementsByBlock[]> {
+    throw new NotImplementedError()
+  }
+
+  getDeposits(_input: GetEventsInput): Promise<MovementsByBlock[]> {
+    throw new NotImplementedError()
+  }
```
jpsains, line 340: 🤦 ok this is annoying needing to throw here

jpsains: maybe we create a "simple-ish borrow adapter" and hide that underthehood

src/core/utils/filters.ts
```diff
@@ -1,18 +1,22 @@
...
 }
 
 export const filterMapAsync = async <Input, Return>(
-  array: Input[],
+  array: readonly Input[],
   callback: (
     value: Input,
     index: number,
-    array: Input[],
+    array: readonly Input[],
```
jpsains, line 19: nice

jpsains (approved):
lgtm

PR #183, opened 2024-03-07
feat: convex

jpsains:


src/adapters/convex/products/staking/metadata/ethereum.metadata.json
```diff
@@ -1,4 +1,28 @@
...
+    "rewardTokens": [
+      {
+        "address": "0xD533a949740bb3306d119CC777fa900bA034cd52",
+        "name": "Curve DAO Token",
+        "symbol": "CRV",
+        "decimals": 18
+      }
+    ],
+    "extraRewardTokens": [],
+    "extraRewardTokenManagers": []
```
bergarces, line 24: nit: If `extraRewardTokens` and `extraRewardTokenManagers` are supposed to be the same size, it'd have been better if they were the same array (in which each item has two fields, the token metadata and the token manager).

Otherwise we'd have to use `!` in typescript for array access or do an `undefined` check.

In any case, not worth changing it.

src/types/adapter.ts
```diff
@@ -164,6 +164,11 @@ export type ProtocolDetails = {
   assetDetails: AssetDetails
 }
 
+export interface GetPositionsInputWithMandatoryTokenAddresses
```
bergarces, line 167: nit: `GetPositionsInputWithTokenAddresses` just to shorten it a bit whilst keeping the meaning.

src/core/utils/getTokenMetadata.ts
```diff
@@ -71,22 +66,30 @@ async function getOnChainTokenMetadata(
...
       { tokenAddress, chainId, errorMessage },
       'Failed to fetch token metadata on-chain',
     )
-    return undefined
+    return {
+      address: getAddress(await tokenContract.getAddress()),
+      name: DEFAULT_STRING,
+      symbol: DEFAULT_STRING,
+      decimals: DEFAULT_DECIMAL,
+    }
```
bergarces, line 92: Problem here. On line 51 of this same file we check if the result from here is undefined and throw an error. That cannot happen any longer after this change.

If we are going to change that behaviour we have to:
- Remove the code that checks if this result is undefined, as it's no longer the case
- Make sure that logic is not needed elsewhere.

My main worry about doing this, is that if there's a hiccup when we are generating metadata, this will just simply create tokens with empty addresses.

src/adapters/convex/products/sidechain-staking/convexSidechainStakingAdapter.ts
```diff
@@ -0,0 +1,258 @@
...
+          extraRewardTokens,
+          extraRewardTokenManagers,
+        }
+      }),
+    )
+
+    return metadata
+  }
+
+  private async getExtraRewardTokensMetadata(
```
bergarces, line 101: This should have gone in the previous PR, but shouldn't these `extra*` methods be private or protected? If they are not going to be called from `DefiProvider` they should not be public.

src/adapters/convex/products/staking/convexStakingAdapter.ts
```diff
@@ -319,7 +305,7 @@ export class ConvexStakingAdapter
       extraRewardTokenManagers,
     } = await this.fetchPoolMetadata(protocolTokenAddress)
 
-    const responsePromises = protocolRewardTokens?.map(
+    const responsePromises = protocolRewardTokens!.map(
```
bergarces, line 308: If this is an internal call (I can only see it being called within `LpStakingProtocolAdapter`), we probably want to make this property non-optional so that we don't have to assert.

src/adapters/convex/products/sidechain-staking/convexSidechainStakingAdapter.ts
```diff
@@ -0,0 +1,258 @@
...
+    await Promise.all(
+      Array.from({ length: Number(length) }, async (_, i) => {
+        const convexData = await convexFactory.poolInfo(i)
+
+        const [
+          convexToken,
+          underlyingToken,
+          [extraRewardTokens, extraRewardTokenManagers],
+        ] = await Promise.all([
+          getTokenMetadata(convexData.rewards, this.chainId, this.provider), // convex staking contract is missing name, symbol, decimal
```
bergarces, line 80: image

src/adapters/convex/products/sidechain-staking/convexSidechainStakingAdapter.ts
jpsains: Yep they are private. I missed in one of the MRs but they are private now, thanks

PR #199, opened 2024-03-25
creates support  map

bergarces:
Creates new method that returns a support map for all the protocols, products and chains.

`npm run support`

Options
`-p <protocols>`
`-c <chains>`

image

### DIFF
--- a/src/defiProvider.ts
+++ b/src/defiProvider.ts
@@ -571,6 +571,13 @@ export class DefiProvider {
     })
   }
 
+  getSupport(input: {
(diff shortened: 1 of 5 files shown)


src/defiProvider.ts
```diff
@@ -571,6 +571,13 @@ export class DefiProvider {
     })
   }
 
+  getSupport(input: {
```
bergarces, line 574: Suggestions for a better name appreciated.

jpsains (approved):

PR #200, opened 2024-03-26
MMI-4061-recursive-underlying-tokens-for-tvl

bergarces:


src/core/utils/resolveUnderlying.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export async function resolveUnderlyings(
+  adapter: IProtocolAdapter,
+  blockNumber: number | undefined,
+  tokens: Token[],
+  updateUnderlyingToken: (
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    underlyingToken: any,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    protocolToken: any,
```
bergarces, line 23: There has to be a way of sorting these `any` with generics, but I'm hitting a brick wall right now, so I need to give it a few hours to see if the penny drops.

src/defiProvider.ts
```diff
@@ -120,6 +121,17 @@ export class DefiProvider {
...
       })
 
+      await resolveUnderlyings(
+        adapter,
+        blockNumber,
+        protocolPositions,
+        (underlyingToken, protocolToken, underlyingRateRaw) => {
+          underlyingToken.balanceRaw =
+            (protocolToken.balanceRaw * underlyingRateRaw) /
+            10n ** BigInt(protocolToken.decimals)
```
bergarces, line 131: This method is different depending on the field that needs being added (`balanceRaw` for positions and movements, and `totalValueRaw` for TVL).

One possible alternative here, since the calculation seems to be always the same, is to pass instead the field name that needs to be updated as a string (so it can be accessed as an index). I'll see how that looks, but it can cause more type problems.

src/core/getProfits.ts
```diff
@@ -85,6 +125,19 @@ export async function getProfits({
         protocolTokenAddresses: Object.keys(endPositionValues), // endPositionValues is indexed by tokenId ?? protocolTokenAddress
         tokenIds: Object.keys(endPositionValues), // endPositionValues is indexed by tokenId ?? protocolTokenAddress
       })
+      .then(async (result) => {
+        await resolveUnderlyings(
+          adapter,
+          fromBlock,
+          result,
+          (underlyingToken, protocolToken, underlyingRateRaw) => {
```
jpsains: I think in this file the function is always the same yeah? Do you mind extracting the function?

e.g. rateCalculation

jpsains (approved):

PR #207, opened 2024-04-03
feat: cur voting escrow

jpsains:
- vesting maturity date appended to protocol token name 
- created curve voting escrow abstract class 
- stopped short of turning this into a template as stargate re-wrote this voting escrow in solidity so now clear to me atm what parts are reusable 
- apologies for all other metadata updates but left them in (let me know if you want them removed)


src/adapters/stargate/products/voting-escrow/metadata/ethereum.protocol-token.json
```diff
@@ -0,0 +1,52 @@
...
+      "symbol": "USDC",
+      "decimals": 6
+    }
+  },
+  {
+    "address": "0xAF667811A7eDcD5B0066CD4cA0da51637DB76D09",
+    "tag_id": "is_reward_manager",
+    "value": true
+  }
+]
```
jpsains, line 52: Bit of a spike to try out the "tag" format.

src/core/adapters/votingEscrow.ts
```diff
@@ -0,0 +1,380 @@
...
+   * Creates a reward manager tag
+   */
+  createRewardManagerMetadata(address: string): Metadata {
+    return {
+      address: address,
+      tag_id: TagIds.isRewardManager,
+      value: true,
+    }
+  }
+
```
jpsains, line 175: Left the utils in this adapter for now......if we think the solution has some appeal these could live in a MetadataBuilder class

bergarces (approved):
Very nice template solution for this sort of token.

src/adapters/stargate/products/voting-escrow/metadata/ethereum.protocol-token.json
bergarces: Let's start with this, we don't need to commit to anything specific yet as we can always regenerate it.

PR #209, opened 2024-04-04
feat: angle protocol adapter

0xtekgrinder:
## **Description**

Add the savings solution adaptor of angle protocol to mmi
documentation can be found [here](https://docs.angle.money/savings/savings)

- [X] A brief description of the protocol and adapters is added to the PR
- [X] Files outside the protocol folder are not being edited and, if they are, it's clearly explained why in the PR
- [X] `update me` comments are removed
- [X] Contracts used are verified for that chain block explorer
- [X] Test cases with a block number have been added to the `testCases.ts` file for every relevant method that has been implemented
  - [X] positions
  - [X] prices
  - [X] deposits
  - [X] withdrawals
  - [X] profits
  - [X] tvl
  - [X] apr
  - [X] apy
- [X] `getAddress` from `ethers`
  - [X] Is used to parse hardcoded addresses
  - [X] Is used to parse addresses that come from contract calls when it is not clear they'll be in checksum format
  - [X] It is NOT used to parse addresses from input methods
  - [X] It is NOT used to parse addresses from metadata
- [X] If the adapters requires to fetch static data on-chain (e.g. pool ids, token metadata, etc)
  - [X] The adapter implements `IMetadataBuilder`
  - [X] The `buildMetadata` method is implemented with the `@CacheToFile` decorator
  - [X] All the static data is stored within the metadata JSON file


0xtekgrinder:
We still have issues regarding metadata that is being cached which ends up having agEUR as symbol instead of EURA.

Also when testing in my machine the fetch of the price wasn't working

jpsains:
Hey @0xtekgrinder,

Great work mate.

I didnt have access to push some small changes to your fork so merged your adapter in a new branch.

image

How did you find it, any feedback is much appreciated :)  

Oh and we will fix the symbol (we have some old Arbitrum and Mainnet symbols coming from a json file, we are going to drop  those old json files in new MR )

PR #213, opened 2024-04-09
Feature: velocore adapter

velocore:
## **Description**

Velocore is a veDEX with real-time voting.

Website: https://velocore.xyz
App: https://linea.velocore.xyz
Docs: https://docs.velocore.xyz
Contracts: https://docs.velocore.xyz/security-and-contract-address/linea-contracts

We have a different architecture to the Uniswap forks. 
For example, we have a singleton vault structure for all kinds of interactions, erc-1155 LPs for stable pools, and a gauge contract for each LP.
So we implemented VelocorePoolAdapter based on the IProtocolAdapter class rather than SimplePoolAdapter.

Please note that we used gauge address instead of pool address as the key of the metadata object.
Because of our structure, a stable pool can have multiple LPs in a pool, each LP is an erc-1155 token, distinguished by tokenId, and each LP has one gauge associated with it.
Therefore, we used the gauge address as the key of the metadata object for a one-to-one mapping of protocolToken and key.

The rest of the VelocorePoolAdapter is nothing special. We used the Lens contract to get the data for the gauges and pools.

  - [ ] positions
  - [ ] prices
  - [ ] deposits
  - [ ] withdrawals
  - [ ] profits
  - [ ] tvl
  - [ ] apr
  - [ ] apy


src/defiProvider.ts
```diff
@@ -211,31 +212,44 @@ export class DefiProvider {
...
 
       const tokens = await Promise.all(
-        protocolTokens.map(async ({ address }) => {
+        protocolTokens.map(async ({ address, tokenId }) => {
           const startTime = Date.now()
 
           const unwrap = await adapter.unwrap({
             protocolTokenAddress: getAddress(address),
             blockNumber,
+            tokenId,
```
velocore, line 252: We found that the defiProvider.unwrap function doesn't accept filterTokenId, even though the TestCase type has the filterTokenId property. 
https://github.com/consensys-vertical-apps/mmi-defi-adapters/blob/1c5bdabdceb5a77371d18e0beec7db49494ece87/src/types/testCase.ts#L73-L78

Also, the adapter.unwrap function has tokenId as  a parameter.
https://github.com/consensys-vertical-apps/mmi-defi-adapters/blob/1c5bdabdceb5a77371d18e0beec7db49494ece87/src/types/adapter.ts#L56-L70

We need tokenId to handle our stable pool LPs (erc-1155) correctly, so we decided to add tokenId as params.

It is outside of our adapter implementation, so please let me know if it is not acceptable.

PR #221, opened 2024-04-22
Library-ui

bergarces:


.gitignore
bergarces, line 1: I've picked a more js-oriented example for gitignore, as it was ignoring the `packages` folder due to some .NET rules.

PR #222, opened 2024-04-23
changes for monorepo

bergarces:
- Use `dist` for output, more standard than `lib/main`
- Use a `.gitignore` more js-centric so that it doesn't ignore `packages` folder.
- Add linting rules to eslintrc.js file
- Add new option to enable Infura Failover (enabled by default)
- Update docs
- Added missing types to `CalculationData`


.env.example
```diff
@@ -1,4 +1,4 @@
-DEFI_ADAPTERS_LOG_LEVEL=debug
+DEFI_ADAPTERS_LOG_LEVEL=info
```
bergarces, line 1: Default log level moved to info because using debug, when running new adapter command, it'll ask the first question and then dump all sort of debug lines in between.

src/types/adapter.ts
```diff
@@ -309,6 +309,8 @@ export interface PositionProfits extends Erc20Metadata, UnderlyingProfitValues {
 export interface CalculationData {
   withdrawals: number
   deposits: number
+  repays: number
+  borrows: number
```
bergarces, line 313: This was missing from the type, but it's being added in the object.

tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "resolveJsonModule": true,
     "strict": true,
     "noUncheckedIndexedAccess": true,
+    "skipLibCheck": true,
```
bergarces, line 15: This should always be true, if there's an issue with the types of a library in `node_modules`, we don't want it to stop us from building.

jpsains (approved):

PR #223, opened 2024-04-23
Move-to-monorepo

bergarces:
- Creates monorepo structure
- Adds library code as package `adapters-library`
- Creates `dev-ui` package for the UI.


.env.example
```diff
@@ -1,4 +1,4 @@
-DEFI_ADAPTERS_LOG_LEVEL=debug
+DEFI_ADAPTERS_LOG_LEVEL=info
```
bergarces, line 1: This was interfering with the new-adapters cCLI command. It would dump debug lines after the first question is asked and before I had a chance to answer.

.github/workflows/publish.yaml
bergarces, line 1: I'm not 100% sure if this is going to work.

I'm basically going to run `npm version` in the library package and use the tag for the repo itself. The version in the main package and the ui do not matter.

packages/dev-ui/vite.config.ts
```diff
@@ -0,0 +1,56 @@
...
+  'DEFI_ADAPTERS_USE_FAILOVER',
+]
+
+const projectEnvs = envs.reduce(
+  (acc, env) => {
+    acc[`process.env.${env}`] = JSON.stringify(process.env[env])
+    return acc
+  },
+  {} as Record<string, string>,
+)
```
bergarces, line 37: Magic code that forces our envs into vite, which for security reasons otherwise just takes envs that start with `VITE_*`. This means we need to add new envs to this list (or to do something clever such as reading the .env file and creating the list automatically).

jpsains (dismissed):

jpsains (approved):

PR #226, opened 2024-04-29
feat: cli fixes

jpsains:


packages/adapters-library/src/scripts/templates/templates.ts
```diff
@@ -0,0 +1,21 @@
...
+  ['CompoundV2']: [
+    compoundV2SupplyMarketForkAdapterTemplate,
+    compoundV2BorrowMarketForkAdapterTemplate,
+  ],
+  ['VotingEscrowAdapter (like curve and stargate voting escrow)']: [
+    votingEscrowAdapterTemplate,
+  ],
+  ['WriteOnlyDeFiAdapter (supports only create transaction params, no getPositions features)']:
+    [writeOnlyDeFiAdapter],
+}
```
jpsains, line 21: Now an array of templates meaning we can generate ore than one by default

PR #235, opened 2024-05-02
await getSupport method

bergarces:


packages/adapters-library/src/scripts/buildTypes.ts
```diff
@@ -99,15 +99,8 @@ async function buildSchemas() {
 
     for (const adapterSupport of protocolAdapters) {
       const productId = adapterSupport.protocolDetails.productId
-      const adapterFilePath = path.resolve(
-        `./packages/adapters-library/dist/adapters/${protocolId}/products/${productId}/${lowerFirst(
-          protocolKey,
-        )}${pascalCase(productId)}Adapter`,
-      )
-
-      const { WriteActionInputs } = await import(adapterFilePath)
```
bergarces, line 108: No need to do this as now the `getSupport` endpoint returns the array of write actions.

PR #245, opened 2024-05-23
Mmi 4383 create integration test without filters

bergarces:


packages/adapters-library/src/core/provider/MulticallQueue.ts
```diff
@@ -121,7 +121,10 @@ export class MulticallQueue {
             flushTimeoutMs: this.flushTimeoutMs,
             maxBatchSize: this.maxBatchSize,
             batchSize,
-            error: error?.info?.error,
+            error: {
+              code: error.code,
+              message: error.message,
+            },
```
bergarces, line 127: It now logs the correct error.

jpsains (dismissed):

PR #247, opened 2024-05-29
Performance script and config changes

bergarces:
- Added performance script to run multiple iterations of get positions for performance results
- Changes so that unwrap only gets called once
- Minor config changes (batch size and stable block for arbitrum)
- More logs


packages/adapters-library/src/core/provider/CustomMulticallJsonRpcProvider.ts
```diff
@@ -39,8 +41,8 @@ export class CustomMulticallJsonRpcProvider extends CustomJsonRpcProvider {
     })
     this.multicallQueue = new MulticallQueue({
       fetchRequest,
-      maxBatchSize: 100,
-      flushTimeoutMs: 1000,
+      maxBatchSize,
+      flushTimeoutMs: 100,
```
bergarces, line 45: Reduced flush to 0.1s, which is what we had before. We shouldn't have to wait a whole second before new calls are received as that could delay responses by a lot.

.env.example
```diff
@@ -11,7 +11,7 @@ DEFI_ADAPTERS_PROVIDER_ARBITRUM=https://arbitrum-mainnet.infura.io/v3/abafec30d6
 DEFI_ADAPTERS_PROVIDER_AVALANCHE=https://avalanche-mainnet.infura.io/v3/abafec30d6aa45ffa0c763b5552a2d02
 DEFI_ADAPTERS_PROVIDER_LINEA=https://linea-mainnet.infura.io/v3/abafec30d6aa45ffa0c763b5552a2d02
 DEFI_ADAPTERS_ENABLE_USD_PRICES_FOR_POSITIONS=true
-DEFI_ADAPTERS_RPC_CALL_TIMEOUT_IN_MS=20000
+DEFI_ADAPTERS_RPC_CALL_TIMEOUT_IN_MS=10000
```
bergarces, line 14: This is already the default in the config. Better not to override it here.

packages/adapters-library/src/adapters/curve/tests/snapshots/ethereum.positions.position-2.json
```diff
@@ -84,6 +85,17 @@
...
+              "name": "Curve DAO Token",
+              "symbol": "CRV",
+              "decimals": 18,
+              "iconUrl": "https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xD533a949740bb3306d119CC777fa900bA034cd52/logo.png",
+              "address": "0xD533a949740bb3306d119CC777fa900bA034cd52",
+              "balanceRaw": "97118831066726817n",
+              "type": "underlying-claimable",
+              "balance": 0.09711883106672682,
+              "price": 0.5718285537644536
+            },
```
bergarces, line 98: It's the same value, it's just been moved here because now we get rewards before unwrapping.

packages/adapters-library/src/config.ts
```diff
@@ -116,6 +109,76 @@ const ConfigSchema = z
...
+          .number()
+          .default(
+            parseNumberEnv(process.env.DEFI_ADAPTERS_MAX_BATCH_SIZE_BASE, 100),
+          ),
+        arbitrum: z
+          .number()
+          .default(
+            parseNumberEnv(
+              process.env.DEFI_ADAPTERS_MAX_BATCH_SIZE_ARBITRUM,
+              5,
```
bergarces, line 161: Default of 5 for arbitrum.

packages/adapters-library/src/core/provider/CustomJsonRpcProvider.ts
```diff
@@ -81,9 +80,17 @@ export class CustomJsonRpcProvider extends JsonRpcProvider {
...
-      currentBlockNumber - (AVERAGE_BLOCKS_PER_10_MINUTES[this.chainId] ?? 0) // default to 0 to avoid failures
-    return blockNumberTenMinsAgo
+    const blocksToMoveBack = (() => {
+      switch (this.chainId) {
+        case Chain.Arbitrum:
+          return AVERAGE_BLOCKS_PER_10_MINUTES[this.chainId] / 4
+        default:
+          return AVERAGE_BLOCKS_PER_10_MINUTES[this.chainId]
+      }
+    })()
```
bergarces, line 90: Changed stable block for arbitrum to 2.5m.

Not massively better, but spamming postman gave me the impression that the closer the block, the less chance of it taking an unreasonable amount of time.

packages/adapters-library/src/core/utils/unwrap.ts
```diff
@@ -87,7 +108,8 @@ async function fetchUnwrapExchangeRates(
       !(
         error instanceof
         ProtocolSmartContractNotDeployedAtRequestedBlockNumberError
-      )
+      ) &&
+      !(error instanceof NotImplementedError)
```
bergarces, line 112: I had to add this because with convex, it returns a protocol token with no underlyings other than claimable rewards. With the new logic that meant it would be unwrapped, but that was causing an error due to the convex curve wrapper adapter not implementing unwrap.

packages/adapters-library/src/core/utils/unwrap.ts
```diff
@@ -47,8 +60,16 @@ export async function unwrap(
...
-      (unwrappedTokenExchangeRate) => {
+    if (!unwrapExchangeRates?.tokens) {
+      return
+    }
+
+    if (!token.tokens) {
+      token.tokens = []
+    }
+
+    token.tokens.push(
```
bergarces, line 71: Needs to push instead of replace, as otherwise it removes the rewards which have already been added.

packages/adapters-library/src/core/provider/CustomMulticallJsonRpcProvider.ts
jpsains: I think I accidentally change this on my last MR

packages/adapters-library/src/config.ts
jpsains: cool

packages/adapters-library/src/core/utils/unwrap.ts
jpsains: ah i see...perhaps we need to implement unwrap, one for another day

jpsains (approved):

PR #251, opened 2024-06-05
remove morpho empty test

bergarces:
Remove empty test that has already a good alternative.


packages/adapters-library/src/config.ts
```diff
@@ -124,7 +124,7 @@ const ConfigSchema = z
           .default(
             parseNumberEnv(
               process.env.DEFI_ADAPTERS_MAX_BATCH_SIZE_OPTIMISM,
-              100,
+              20,
```
bergarces, line 127: Sonny integration tests keep timing out when sending batches. I ran a few performance tests for optimism and 20 seems to be the sweet spot.

jpsains (approved):

PR #265, opened 2024-06-19
Pendle

jpsains:


packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
```diff
@@ -0,0 +1,305 @@
...
+          })
+          .catch((e) => {
+            return oracle.getYtToSyRate(metadata!.marketAddress, 1800, {
+              blockTag: blockNumber,
+            })
+          })
+
+        break
+      case 'lp':
+        rate = await oracle
```
bergarces, line 109: You don't need a default here, the type must be one of those 4.

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
```diff
@@ -0,0 +1,305 @@
...
+    const metadata = (await this.buildMetadata())[protocolTokenAddress]
+    const underlyingToken = metadata!.underlyingToken
+
+    const oracle = OraclePyYtLp__factory.connect(
+      '0x9a9fa8338dd5e5b2188006f1cd2ef26d921650c2',
+      this.provider,
+    )
+
+    let rate: bigint
+    switch (metadata!.type) {
```
bergarces, line 83: If possible, it'd be nice to have an explanation or a reference about:
- What is this type
- What is that 900 that is passed to some of those methods

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
```diff
@@ -0,0 +1,305 @@
...
+    this.adaptersController = adaptersController
+    this.helpers = helpers
+  }
+
+  async unwrap({
+    blockNumber,
+    protocolTokenAddress,
+    tokenId,
+  }: UnwrapInput): Promise<UnwrapExchangeRate> {
+    const metadata = (await this.buildMetadata())[protocolTokenAddress]
```
bergarces, line 74: You can use `fetchPoolMetadata` and avoid `!`

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
```diff
@@ -0,0 +1,305 @@
...
+          .catch((e) => {
+            return oracle.getYtToSyRate(metadata!.marketAddress, 1800, {
+              blockTag: blockNumber,
+            })
+          })
+
+        break
+      case 'sy': {
+        rate = 10n ** 18n
+
```
bergarces, line 122: Doesn't need the second `!`, and the first one can be avoided to if it's fetch with `fetchPoolMetadata`.

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
```diff
@@ -0,0 +1,305 @@
...
+    const metadata: Metadata = {}
+
+    resp.results.map((value) => {
+      const market = getAddress(value.address)
+
+      const pt: Erc20Metadata = {
+        address: getAddress(value.pt.address),
+        name: value.pt.name,
+        symbol: value.pt.symbol,
+        decimals: value.pt.decimals,
```
bergarces, line 174: `value.underlyingAsset` cannot be undefined according to their type, and none of the properties need `!`.

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
```diff
@@ -0,0 +1,305 @@
...
+      }
+      const yt: Erc20Metadata = {
+        address: getAddress(value.yt.address),
+        name: value.yt.name,
+        symbol: value.yt.symbol,
```
bergarces, line 179: Same as before, `value.sy` is always defined.

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
jpsains: yes, long story, was switching out tokens from their api, the shape of their metadata was not obvious, must have had a token there that was optional, good spot

packages/adapters-library/src/adapters/pendle/products/market/pendleMarketAdapter.ts
jpsains: not sure i get your point here?

bergarces: I meant value.yt didn't need the !. Sorted now anyway.

bergarces (approved):

PR #272, opened 2024-07-01
feat: pendle update

jpsains:
- break pendle adapter down into separate adapters
- remove the default to 0 when eth_call request fails


packages/adapters-library/src/adapters/pendle/tests/snapshots/ethereum.prices.4.json
```diff
@@ -33,6 +33,57 @@
...
+      "chainId": 1,
+      "productId": "principle-token",
+      "success": false,
+      "error": {
+        "message": "Protocol token pool not found",
+        "details": {
+          "name": "Error"
+        }
+      }
+    },
```
jpsains, line 53: is expected as cant filter by poduct id

bergarces (approved):
Much better like this.

PR #284, opened 2024-07-26
feat: new metadata types

jpsains:
* apply new build-metadata types
* update stargate 
* includes switching stargate over to new adapter types 
* ensure decorator is called when added to getProtocolTokens


packages/adapters-library/src/scripts/buildMetadata.ts
```diff
@@ -93,6 +81,50 @@ export function buildMetadata(
                 }
               | undefined
 
+            if (adapter.adapterSettings.version === 2) {
```
jpsains, line 84: open to suggestions here...ideally we want to check if .getProtocolTokens has cache decorator applied...or maybe we dont need to check and we just call getProtocolTokens(true)

packages/adapters-library/src/types/IProtocolAdapter.ts
```diff
@@ -20,6 +20,24 @@ import type {
...
+
+export type JsonMetadata = Record<string, Json>
+
+export type ProtocolToken<
+  // biome-ignore lint/complexity/noBannedTypes: <explanation>
+  AdditionalMetadata extends JsonMetadata = {},
+> = Erc20Metadata & {
+  underlyingTokens?: Erc20Metadata[]
+  tokenId?: string
+} & AdditionalMetadata
```
bergarces, line 39: As well as having `underlyingTokens` and `tokenId` as optional, we could have two different types that extend from ProtocolToken that have them concretely. Something like `ProtocolTokenWithUnderlyings` and `ProtocolTokenNFT`.

Otherwise we have to keep specifying them as `AdditionalMetadata`, even though it's not really additional, just optional.

packages/adapters-library/src/types/IProtocolAdapter.ts
```diff
@@ -20,6 +20,24 @@ import type {
...
+  | number
+  | string
+  | Json[]
+  | { [prop: string]: Json }
+
+export type JsonMetadata = Record<string, Json>
+
+export type ProtocolToken<
+  // biome-ignore lint/complexity/noBannedTypes: <explanation>
+  AdditionalMetadata extends JsonMetadata = {},
```
bergarces, line 35: I think biome is messing up here, we are not using `{}` as a type, but as a default value.

packages/adapters-library/src/types/adapter.ts
```diff
@@ -108,6 +108,7 @@ export type GetEventsInput = {
 export type AdapterSettings = {
   enablePositionDetectionByProtocolTokenTransfer: boolean
   includeInUnwrap: boolean
+  version: number
```
bergarces, line 111: :+1:

packages/adapters-library/src/types/IProtocolAdapter.ts
bergarces: On the other hand, we don't have to make that call now. What we store in itself won't change one way or another. So this is not a blocker.

packages/adapters-library/src/scripts/buildMetadata.ts
bergarces: Passing extra parameters won't make it fail if it's not used, so it's fine to do it this way.

We cannot directly tell if `getProtocolTokens` is decorated, so I think this is good enough.

bergarces (approved):

bergarces (approved):

PR #285, opened 2024-07-26
interceptor

bergarces:


packages/adapters-library/src/core/provider/MulticallQueue.ts
```diff
@@ -59,6 +59,9 @@ export class MulticallQueue {
     this.provider = new ethers.JsonRpcProvider(
       fetchRequest,
       Network.from(chainId),
+      {
+        staticNetwork: Network.from(chainId),
+      },
```
bergarces, line 64: This was causing us to spam infura on every `eth_call` with an `eth_chainId`.

packages/adapters-library/src/adapters/ethena/tests/snapshots/ethereum.deposits.json
```diff
@@ -47,5 +47,395 @@
...
+            "data": "0x82ad56cb0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000009d39a5de30e57443bff2a8307a4256c8797a349700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000002470a082310000000000000000000000003f843189280a4379eb12b928afd5d96df807667900000000000000000000000000000000000000000000000000000000"
+          },
+          "0x12dba9c"
+        ]
+      },
+      "response": "0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000682a6ecd45edf435efda8"
+    },
```
bergarces, line 64: For the record, we do not need to store the request data, just the response. I was only storing the request for debugging purposes but now I'm wondering if we should remove it or not, as it's what allowed me to detect the unnecessary Lido calls even for non-Lido tests.

packages/adapters-library/src/scripts/rpcInterceptor.ts
```diff
@@ -0,0 +1,119 @@
...
+    .update(url + JSON.stringify(requestBody))
+    .digest('hex')
+}
+
+function slimRequest({ method, params }: RpcRequest) {
+  return {
+    method,
+    params,
+  }
+}
```
bergarces, line 23: Needs to strip `id` as it's an arbitrary number on every execution, so it cannot be part of the hash.

packages/adapters-library/src/scripts/rpcInterceptor.ts
```diff
@@ -0,0 +1,119 @@
...
+  method: string
+  params: Json[]
+  id: number
+}
+
+function createKey(url: string, requestBody: Json) {
+  return createHash('md5')
+    .update(url + JSON.stringify(requestBody))
+    .digest('hex')
+}
```
bergarces, line 16: Added Url to the hash just in case two different chains have the same parameters on a request.

packages/adapters-library/src/core/provider/MulticallQueue.ts
jpsains: nice save

packages/adapters-library/src/adapters/ethena/tests/snapshots/ethereum.deposits.json
jpsains: ignore my env var suggestion, i think that would be overkill

bergarces: Let's just strip them for now.

If we need to debug something in future it's easy to temporarily modify it.

jpsains (approved):

packages/adapters-library/src/adapters/iziswap/tests/testCases.ts
```diff
@@ -45,18 +47,6 @@ export const testCases: TestCase[] = [
...
-    method: 'withdrawals',
-    input: {
-      userAddress: '0x4F21a0Af178d9EB793c2D94d9d88910e974a98fD',
-      fromBlock: 34127856,
-      toBlock: 34127858,
-      protocolTokenAddress: '0xBF55ef05412f1528DbD96ED9E7181f87d8C9F453',
-      productId: 'pool',
-      tokenId: '85',
-    },
-  },
```
bergarces, line 59: No withdrawals for iziswap on BSC recent enough for the request to work with our rpc provider.

jpsains (approved):

PR #289, opened 2024-07-31
feat: step 1

jpsains:
* converts aave over to new adapters (incase people copy aave adapter again)
* converts SimplePoolAdapter to use the new helpers


packages/adapters-library/src/adapters/aave-v2/common/aaveBasePoolAdapter.ts
```diff
@@ -181,7 +262,7 @@ export abstract class AaveBasePoolAdapter
     reserveData: Awaited<ReturnType<ProtocolDataProvider['getReserveData']>>,
   ): bigint
 
-  private async fetchPoolMetadata(protocolTokenAddress: string) {
+  async fetchPoolMetadata(protocolTokenAddress: string) {
```
bergarces, line 265: Why is this no longer private? I cannot find any instance of it being called outside of the class context.

packages/adapters-library/src/adapters/aave-v2/products/a-token/aaveV2ATokenAdapter.ts
```diff
@@ -1,15 +1,26 @@
...
+  GetPositionsInput,
+  GetTotalValueLockedInput,
+  MovementsByBlock,
   PositionType,
+  ProtocolAdapterParams,
   ProtocolDetails,
+  ProtocolPosition,
+  ProtocolTokenTvl,
+  UnwrapExchangeRate,
+  UnwrapInput,
```
bergarces, line 17: nit: All these imports are not required.

bergarces: Ignore this. It's better to go ahead. I'll see if I can introduce it as a mandatory lint error and correct them automatically, as other files are affected too.

bergarces (approved):

PR #291, opened 2024-08-01
Disable-deri-bsc

bergarces:


bergarces:
> why are we disabling?

Deri adapter is NFT based and has no way of fetching user positions in the contract. It needs to be done by fetching all the user events, which is impossible in BSC.

PR #296, opened 2024-08-16
MMI-5380-update-blank-template-to-build-metadata-inside-get-protocol-tokens

bergarces:


packages/adapters-library/src/scripts/templates/blankAdapterTemplate.ts
```diff
@@ -39,35 +36,30 @@ export function blankAdapterTemplate({
...
-      string,
-      {
-        protocolToken: Erc20Metadata
-        underlyingTokens: Erc20Metadata[]
-      }
-    >
-    
-    export class ADAPTER_CLASS_NAME implements IProtocolAdapter, IMetadataBuilder {
+
+    type AdditionalMetadata = { underlyingTokens: Erc20Metadata[] }
```
bergarces, line 45: I really think that `underlyingTokens` shouldn't `AdditionalMetadata`. I can't think of a single use case when we need metadata and this is not needed.

jpsains: as discussed needed while we have 2 versions of the metadata types

jpsains (approved):

PR #301, opened 2024-08-26
Mmi 5327 solv yield market profits

xavier-brochard:
## **Description**

- Solv Yield Market adapter:
  - Refactor the Solv Yield Market adapter with a metadata-oriented config, paving the way for actual metadata generation
  - Add support to deposits and withdrawals

- [x] A brief description of the protocol and adapters is added to the PR
- [x] Files outside the protocol folder are not being edited and, if they are, it's clearly explained why in the PR
- [x] `update me` comments are removed
- [x] Contracts used are verified for that chain block explorer
- [x] Test cases with a block number have been added to the `testCases.ts` file for every relevant method that has been implemented
  - [x] positions
  - [x] profits
  - [ ] prices (unwrap)
  - [x] deposits
  - [x] withdrawals
  - [ ] repays
  - [ ] borrows
  - [ ] tvl
- [x] `getAddress` from `ethers`
  - [x] Is used to parse hardcoded addresses
  - [x] Is used to parse addresses that come from contract calls when it is not clear they'll be in checksum format
  - [x] It is NOT used to parse addresses from input methods
  - [x] It is NOT used to parse addresses from metadata
- [x] For every adapter that extends `SimplePoolAdapter`
  - [x] `getPositions` is not overwritten and, if it is, it's clearly explained why in the PR
  - [x] `unwrap` is not overwritten and, if it is, it's clearly explained why in the PR
- [x] If the adapters requires to fetch static data on-chain (e.g. pool ids, token metadata, etc)
  - [x] The adapter implements `IMetadataBuilder`
  - [x] The `buildMetadata` method is implemented with the `@CacheToFile` decorator
  - [x] All the static data is stored within the metadata JSON file


packages/adapters-library/src/adapters/solv/products/yield-market/solvYieldMarketAdapter.ts
```diff
@@ -266,159 +277,146 @@ export class SolvYieldMarketAdapter implements IProtocolAdapter {
...
-      shareSlot.toString(),
-    )} | Redemption pending`
+    if (!tokensCount) return []
 
-    const {
-      symbol: underlyingSymbol,
-      decimals: underlyingDecimals,
-      name: underlyingName,
-    } = await this.helpers.getTokenMetadata(currency)
+    const indexes = [...Array(Number.parseInt(tokensCount.toString())).keys()]
```
bergarces, line 299:
```suggestion
    const indexes = [...Array(Number(tokensCount)).keys()]
```
A bit less verbose and equally unsafe way of casting from BigInt to Number.

packages/adapters-library/src/adapters/solv/products/yield-market/solvYieldMarketAdapter.ts
```diff
@@ -266,159 +277,146 @@ export class SolvYieldMarketAdapter implements IProtocolAdapter {
...
    *
    * @example
-   * const openFundShareDelegateAddress = '0x22799daa45209338b7f938edf251bdfd1e6dcb32'
    * const slot = '5310353805259224968786693768403624884928279211848504288200646724372830798580'
-   * computePoolId(sftAddress, slot) // returns '0xe037ef7b5f74bf3c988d8ae8ab06ad34643749ba9d217092297241420d600fce'
+   * computePoolId(slot) // returns '0xe037ef7b5f74bf3c988d8ae8ab06ad34643749ba9d217092297241420d600fce'
    *
    * @see https://github.com/solv-finance/SolvBTC/blob/ef5be00ec22549ac5a323378c2a914166bf0dcc1/contracts/SftWrappedToken.sol#L198
    */
   private computePoolId(slot: bigint): string {
```
bergarces, line 318: Looks like this is no longer being used. Any reason to keep it?

bergarces (approved):
LGTM

I haven't gone deep into QAing the whole deposits/withdraws, as that's simply impossible without learning the whole protocol.

packages/adapters-library/src/adapters/solv/products/yield-market/solvYieldMarketAdapter.ts
xavier-brochard: Yes I'd like to keep it, because it will be used to generate metadata

bergarces (approved):

PR #303, opened 2024-08-27
Unwrap and CI fixes

bergarces:
- Refactors unwrap to solve a bug with duplicate rewards (Convex example)
- Fixes Mendi Finance snapshot error by normalizing zeroes.
- Deletes `RewardAdapter` abstract class that is no longer used and blank template reference.
- Deletes helper methods no longer used.


packages/adapters-library/src/adapters/convex/tests/snapshots/ethereum.positions.2.json
bergarces, line 1: This is the example snapshot with duplicate unwrapped rewards.

jpsains (approved):

PR #304, opened 2024-08-28
MMI-5405-investigate-and-solve-for-missing-product-on-stargate-to-ensure-we-have-full-coverage

bergarces:
- Added missing chains and metadata for Stargate V1 LP pools.
- Added support for LP token staking for V1 and V2


packages/adapters-library/src/adapters/protocols.ts
```diff
@@ -33,6 +33,7 @@ export const Protocol = {
   SparkV1: 'spark-v1',
   StakeWise: 'stakewise',
   Stargate: 'stargate',
+  StargateV2: 'stargate-v2',
```
bergarces, line 36: Do we want `stargate-v2` or do we want to group the new products inside `stargate`?

If we want to align our protocol ids with other providers, I don't think they have a stargate-v2.

packages/adapters-library/src/adapters/stargate-v2/common/staticChainData.ts
```diff
@@ -0,0 +1,107 @@
...
+export const staticChainData: Partial<
+  Record<
+    Chain,
+    {
+      poolAddresses: {
+        poolNativeAddress?: string
+        poolUsdcAddress?: string
+        poolUsdtAddress?: string
+        poolMetisAddress?: string
+        poolMethAddress?: string
```
bergarces, line 13: There's no event or any contract I could find where these contract addresses live. They've been fetched from their docs.

https://stargateprotocol.gitbook.io/stargate/v/v2-developer-docs/technical-reference/mainnet-contracts

packages/adapters-library/src/adapters/morpho-blue/contracts/factories/Metamorphofactory__factory.ts
bergarces, line 1: This started crashing my local build due to the json file not matching with the class name.

packages/adapters-library/src/core/utils/caseConversion.ts
bergarces, line 1: It was driving my nuts that every time we did something like `v2` it will be converted to `v-2`, so I amended the implementation.

packages/adapters-library/src/adapters/stargate/products/farm-v2/stargateFarmV2Adapter.ts
```diff
@@ -0,0 +1,374 @@
...
+    const { stargateStakingAddress } = staticChainDataV2[this.chainId]!
+    const protocolTokens = await this.getProtocolTokens()
+
+    return await filterMapAsync(protocolTokens, async (protocolToken) => {
+      if (
+        protocolTokenAddresses &&
+        !protocolTokenAddresses.includes(protocolToken.address)
+      ) {
+        return undefined
+      }
```
jpsains, line 147: Do you see a future where we can drop this filter? Wondering if the filtering could be done outside....

packages/adapters-library/src/adapters/stargate/products/farm-v2/stargateFarmV2Adapter.ts
```diff
@@ -0,0 +1,374 @@
...
+            symbol: protocolToken.symbol,
+            decimals: protocolToken.decimals,
+            type: TokenType.Underlying,
+            blockNumber: event.blockNumber,
+            balanceRaw: amount,
+          },
+        ],
+      }
+    })
+  }
```
jpsains, line 297: This is not a blocker but i  think you might have been able to use the transfer movements of the protocol token

```
    const movements = await this.helpers.getErc20Movements({
      protocolToken: underlyingLpToken,

      filter: {
        fromBlock,
        toBlock,
        from: userAddress,
        to: protocolTokenAddress,
      },
    })
```

jpsains: We have a helper

jpsains: but please dont bother to update its not worth it

packages/adapters-library/src/adapters/stargate/products/farm/stargateFarmAdapter.ts
```diff
@@ -0,0 +1,454 @@
...
+      const lpStakingContract =
+        protocolToken.lpStakingType === 'LpStaking'
+          ? LpStaking__factory.connect(
+              protocolToken.lpStakingAddress,
+              this.provider,
+            )
+          : LpStakingTime__factory.connect(
+              protocolToken.lpStakingAddress,
+              this.provider,
+            )
```
jpsains, line 200: This is fine for now but what stopped you doing a separate adapter?

packages/adapters-library/src/adapters/stargate/products/farm/stargateFarmAdapter.ts
```diff
@@ -0,0 +1,454 @@
...
+          },
+        ],
+      }
+    })
+  }
+
+  /**
+   * There are no specific events for rewards, which occur as simple transfers whenever a Deposit or Withdrawal is actioned
+   * For this reason, we need to fetch all Deposit and Withdrawal events and then filter out the reward transfers
+   */
```
jpsains, line 345: These are the type of scenarios, among others, are the reason why I dont want us to rule out having an indexer....hence why I want @xavier-brochard to at least explorer the indexer solution for ETH2

packages/adapters-library/src/adapters/stargate/products/farm/stargateFarmAdapter.ts
```diff
@@ -0,0 +1,454 @@
...
+        : LpStakingTime__factory.connect(
+            protocolToken.lpStakingAddress,
+            this.provider,
+          )
+
+    const depositFilter = lpStakingContract.filters.Deposit(
+      userAddress,
+      protocolToken.poolIndex,
+      undefined,
+    )
```
jpsains, line 370: Just to confirm are you saying rewards can be "claimed" even with deposits?

packages/adapters-library/src/core/utils/caseConversion.ts
```diff
@@ -19,6 +20,20 @@ function isKebabCase(input: string) {
...
+      const useHyphen =
+        index &&
+        !(wordSplit[index - 1]!.toLowerCase() === 'v' && /^\d+$/.test(word))
+
+      return result + (useHyphen ? '-' : '') + word.toLowerCase()
+    },
+    '',
+  )
+}
+
```
jpsains, line 36: nice thank you

packages/adapters-library/src/adapters/stargate/tests/snapshots/ethereum.withdrawals.farm-v1.json
```diff
@@ -0,0 +1,365 @@
...
+            "address": "0xAf5191B0De278C7286d6C7CC6ab6BB8A73bA2Cd6",
+            "name": "StargateToken",
+            "symbol": "STG",
+            "decimals": 18,
+            "type": "underlying",
+            "blockNumber": 20661441,
+            "balanceRaw": "29168188478059884180158n",
+            "price": 0.27636857144740007,
+            "balance": 29168.188478059885
+          }
```
jpsains, line 35: is this a claimed rewards? If so is the type correct? I think it should it be underlying-claimable?

packages/adapters-library/src/adapters/stargate/products/farm-v2/stargateFarmV2Adapter.ts
```diff
@@ -0,0 +1,374 @@
...
+          const rewardToken = rewardTokens.find(
+            (rewardToken) => rewardToken.address === rewardTokenAddress,
+          )!
+
+          return {
+            address: rewardToken.address,
+            name: rewardToken.name,
+            symbol: rewardToken.symbol,
+            decimals: rewardToken.decimals,
+            type: TokenType.Underlying,
```
jpsains, line 349: underlying-claimable

packages/adapters-library/src/adapters/protocols.ts
jpsains: I dont know what debank have....lets wait for the ticket

packages/adapters-library/src/adapters/morpho-blue/contracts/factories/Metamorphofactory__factory.ts
jpsains: oh ok, not sure what happened there they must have change the name...good spot

packages/adapters-library/src/core/utils/caseConversion.ts
jpsains: yes nice improvement :)

packages/adapters-library/src/adapters/protocols.ts
bergarces: From their UI, it looks like everything is Stargate. Whether the API is the same or not, we don't know yet. 

I've taken the safe approach and all of them are under Stargate now.

packages/adapters-library/src/adapters/stargate/products/farm-v2/stargateFarmV2Adapter.ts
bergarces: To be honest, I can't.

If we do it outside that means the return function will be something like `Promise<ReturnType | undefined>`, which doesn't make a lot of sense.

Moreover, we use that pattern pretty often internally as well, which means that if we don't get rid of it quickly and we chain the result to another array method, the first thing we'll have to do is check if it's undefined.

It's unfortunate that JS doesn't come with an array method that does that automatically.

packages/adapters-library/src/adapters/stargate/products/farm-v2/stargateFarmV2Adapter.ts
bergarces: 100% worth it if we have a helper. I'll update.

packages/adapters-library/src/adapters/stargate/products/farm/stargateFarmAdapter.ts
bergarces: It was my initial thought, since it looked that some chains used one contract and some the other. So my plan was to make a common adapter and then extend for the very tiny different bits.

Then I saw that some chains (mainnet) used both depending on the pool.

Overall, it's a question of whether these ifs are annoying enough to justify a whole new adapter that will either have almost identical code or an abstraction with then several bits of code in protected functions reimplemented for each type.

No strong opinion either way, it's a grey area.

packages/adapters-library/src/adapters/stargate/products/farm/stargateFarmAdapter.ts
bergarces: Yes, rewards get automatically claimed whenever user deposits or withdraws. They don't seem to have a method to claim them directly.
image

packages/adapters-library/src/adapters/stargate/tests/snapshots/ethereum.withdrawals.farm-v1.json
bergarces: I thought so too. But every other adapter implementing the `getRewardWithdrawals` returns them as underlying, so I used this for consistency.

packages/adapters-library/src/adapters/stargate/products/farm-v2/stargateFarmV2Adapter.ts
bergarces: As mentioned before. I'm using the same type that other `getRewardWithdrawals` implementations (curve and convex) use.

Maybe we need to change them all.

PR #305, opened 2024-08-29
feat: solv fix

jpsains:


xavier-brochard (approved):

bergarces (approved):
Approved, but `getTokenIds` should take `blockNumber: number | undefined`.

Using the `!` only works here because passing `undefined` does not throw, but it's not correct.

PR #314, opened 2024-09-10
feat: Added ZeroLend Adapter for MetaMask

deadshotryker:

  - [ ] positions
  - [ ] profits
  - [ ] prices (unwrap)
  - [ ] deposits
  - [ ] withdrawals
  - [ ] repays
  - [ ] borrows
  - [ ] tvl


bergarces (approved):
Sorry for the delay, I rebased and looks good, so merging.

PR #317, opened 2024-09-13
feat: add EtherFi adapters

xavier-brochard:
## **Description**

Adds 4 adapters to support positions and profits in ether.fi's 
- `eETH`
- `weETH`
- Liquid vaults
- L2 restaking

- [x] A brief description of the protocol and adapters is added to the PR
- [x] Files outside the protocol folder are not being edited and, if they are, it's clearly explained why in the PR
- [x] `update me` comments are removed
- [x] Contracts used are verified for that chain block explorer
- [x] Test cases with a block number have been added to the `testCases.ts` file for every relevant method that has been implemented
  - [x] positions
  - [x] profits
  - [x] prices (unwrap)
  - [ ] deposits
  - [ ] withdrawals
  - [ ] repays
  - [ ] borrows
  - [ ] tvl
- [x] `getAddress` from `ethers`
  - [x] Is used to parse hardcoded addresses
  - [x] Is used to parse addresses that come from contract calls when it is not clear they'll be in checksum format
  - [x] It is NOT used to parse addresses from input methods
  - [x] It is NOT used to parse addresses from metadata
- [x] For every adapter that extends `SimplePoolAdapter`
  - [x] `getPositions` is not overwritten and, if it is, it's clearly explained why in the PR
  - [x] `unwrap` is not overwritten and, if it is, it's clearly explained why in the PR


packages/adapters-library/src/adapters/etherfi/products/liquid/etherFiLiquidAdapter.ts
```diff
@@ -0,0 +1,265 @@
...
+    )
+
+    const protocolTokens = await Promise.all(promises)
+
+    /**
+     * TODO: Token 0x5401b8620E5FB570064CA9114fd1e135fd77D57c is twice in this list, making
+     * the method AdaptersController.processDefaultCase throw an error.
+     *
+     * How should we handle properly?
+     */
```
xavier-brochard, line 167: WYT @jpsains @bergarces ?

packages/adapters-library/src/adapters/etherfi/products/l2/etherFiL2Adapter.ts
```diff
@@ -0,0 +1,227 @@
...
+    const mainnetProvider = this.helpers.allJsonRpcProviders[Chain.Ethereum]
+    const weETHContractMainnet = WeETH__factory.connect(
+      WE_ETH_ADDRESS_MAINNET,
+      mainnetProvider,
+    )
+
+    const sameDateMainnetBlockNumber = await getBlockByDate(
+      mainnetProvider,
+      l2Block.date,
+    )
```
bergarces, line 203: This function call will do an average of log(N*) calls to `provider.getBlock` sequentially (unfortunately there doesn't seem to be a `getBlockByTimestamp` available). We will need to check that is doesn't slow down calling `getPositions` for these two L2 chains a lot.

*: N is the number of blocks on mainnet

packages/adapters-library/src/core/utils/getBlockByDate.ts
```diff
@@ -0,0 +1,37 @@
...
+  let upperBound = latestBlockNumber
+
+  while (lowerBound <= upperBound) {
+    const middle = Math.floor((lowerBound + upperBound) / 2)
+    const middleBlock = await provider.getBlock(middle)
+
+    if (!middleBlock) throw new Error(`Block ${middle} not found`)
+
+    // Found the block with the exact timestamp
+    if (middleBlock.timestamp === targetTimestamp) return middleBlock
```
bergarces, line 26: Is this trying to find a block in mainnet with exactly the same timestamp as an L2 block?

What happens if no block in mainnet has the exact same timestamp? This looks like it would throw.

packages/adapters-library/src/adapters/etherfi/products/l2/etherFiL2Adapter.ts
bergarces: Also nit, but can't we pass `l2Block.timestamp` instead of passing `date` and then converting it to a timestamp.

xavier-brochard: Hum, I tend to consider the type `Date` as the first class citizen for function params as it has better devX than timestamp. Especially here `getBlockByDate` is a generic util that could be re-used in various places.

packages/adapters-library/src/core/utils/getBlockByDate.ts
xavier-brochard: > Is this trying to find a block in mainnet with exactly the same timestamp as an L2 block?

Not the same timestamp, there's close to zero chances that this happens. The function `getBlockByDate` finds the block with **closest** timestamp as passed timestamp, using binary search. 

If no exact match (most probable case), it returns the closest (line 36).

xavier-brochard: If you doubt this works, check the unit test, it covers the "not exactly same timestamp" case.

xavier-brochard: I don't mind renaming the function `getBlockWithDateClosestFrom` if that helps.

bergarces: All good, thanks for clarifying.

packages/adapters-library/src/adapters/etherfi/products/liquid/etherFiLiquidAdapter.ts
```diff
@@ -0,0 +1,265 @@
...
+    const uniques = uniqBy(protocolTokens, 'address')
+
+    return uniques
+  }
```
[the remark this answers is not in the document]
bergarces: I saw that, but I can't see the problem with this.

You don't have the protocol token address until after you have collected the whole array.

packages/adapters-library/src/scripts/helpers.ts
```diff
@@ -31,23 +29,12 @@ export const REAL_ESTATE_TOKEN_METADATA = {
...
-    this.provider = provider
-    this.chainId = chainId
-    this.metadataProvider = metadataProvider
-  }
+  constructor(
+    public readonly provider: CustomJsonRpcProvider,
+    public readonly chainId: Chain,
+    public readonly metadataProvider: IMetadataProvider,
+    public readonly allJsonRpcProviders: Record<Chain, CustomJsonRpcProvider>,
+  ) {}
```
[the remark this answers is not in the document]
bergarces: IMHO, yes, this is better as long as there are no optional values involved. I'm a big fan of this syntax for constructors.

But we haven't used it in the code before, thus why I didn't do it either.

bergarces (approved):

PR #323, opened 2024-09-19
injectable unwrap cache

bergarces:


packages/adapters-library/src/defiProvider.ts
```diff
@@ -77,12 +78,15 @@ export class DefiProvider {
...
   private metadataProviders: Record<Chain, IMetadataProvider>
+  private unwrapCache: IUnwrapCache
 
   constructor(
     config?: DeepPartial<IConfig>,
     metadataProviders?: Record<Chain, IMetadataProvider>,
+    unwrapCacheProvider?: IUnwrapCacheProvider,
   ) {
     this.metadataProviders = metadataProviders ?? buildMetadataProviders()
+    this.unwrapCache = new UnwrapCache(unwrapCacheProvider)
```
bergarces, line 89: No provider by default.

Passing an in-memory cache will result in some snapshots changing due to block differences being in the 10 minute range.

packages/adapters-library/src/unwrapCache.ts
```diff
@@ -0,0 +1,77 @@
...
+    return `T${timestampKey}`
+  }
+
+  const blockKey = roundDownToNearestMultiple(
+    blockNumber,
+    AVERAGE_BLOCKS_PER_10_MINUTES[chainId],
+  )
+
+  return `B${blockKey}`
+}
```
bergarces, line 73: An alternative to not use `B` and `T` prefixes would be to convert the timestamp back to an Date ISO string.

Easy to change.

PR #329, opened 2024-09-25
Mountain protocol adapter

bergarces:


packages/adapters-library/src/core/adapters/SimplePoolAdapter.ts
bergarces, line 1: I've removed it, as this method is not being used at all.

PR #330, opened 2024-09-29
MMI-5551-convert-all-the-old-adapters-to-the-our-metadata-types

bergarces:


packages/adapters-library/src/SQLiteMetadataProvider.ts
```diff
@@ -43,109 +45,106 @@ export type PoolRow = {
...
+    ert.additional_data AS extra_reward_additional_data
+  FROM  adapters a 
+        LEFT JOIN pools p ON a.adapter_id = p.adapter_id
+        LEFT JOIN tokens t ON p.pool_address = t.token_address
+        LEFT JOIN underlying_tokens ut ON p.pool_id = ut.pool_id
+        LEFT JOIN tokens ut_t ON ut.token_address = ut_t.token_address
+        LEFT JOIN reward_tokens rt ON p.pool_id = rt.pool_id
+        LEFT JOIN tokens rt_t ON rt.token_address = rt_t.token_address
+        LEFT JOIN extra_reward_tokens ert ON p.pool_id = ert.pool_id
+        LEFT JOIN tokens ert_t ON ert.token_address = ert_t.token_address;
```
bergarces, line 82: Query now returns everything (takes 60ms and it's only 16K rows, so no memory issues).

It then adds everything to in-memory objects that perform the same as the old json files once loaded.

packages/adapters-library/src/SQLiteMetadataProvider.ts
```diff
@@ -43,109 +45,106 @@ export type PoolRow = {
...
-
-    // Return the pool count from the result
-    return Number(result.pool_count ?? 0)
+          resolve(this.formatDbResponse(allRows))
+        } catch (error) {
+          logger.error({ filename, options }, 'Error fetching database rows')
+          reject(error)
+        }
+      },
+    )
```
bergarces, line 106: We start fetching the data when the provider is instantiated, not after the first query. What we store is a promise, so when adapters try to fetch the data they just need to await it, but it should already be there as it's a very fast query.

packages/adapters-library/src/SQLiteMetadataProvider.ts
```diff
@@ -43,109 +45,106 @@ export type PoolRow = {
...
 
-  private selectProtocolPools(
-    db: DbType,
-    protocolId: string,
-    productId: string,
-  ) {
-    const query = this.selectProtocolPoolsQuery
-
-    const rows = db.prepare(query).all(protocolId, productId) as PoolRow[]
+    return adapterTokens ?? []
```
bergarces, line 120: The `?? []` is for adapters that have `@CacheToDb` but no pools. There are a few and we probably want to look at them afterwards (the CLI command shows which ones they are).

packages/adapters-library/src/SQLiteMetadataProvider.ts
```diff
@@ -229,6 +221,57 @@ export class SQLiteMetadataProvider implements IMetadataProvider {
...
+
+    return adaptersMap
+  }
+
+  private adapterKey(protocolId: string, productId: string) {
+    return `${protocolId}#${productId}`
   }
 }
+
+export function buildMetadataProviders(
```
bergarces, line 236: Moved these two functions here from `defiProvider`. I've made a small change that allows us to just pass the the filepath and db options from the lambdas. No need to export SQLiteMetadataProvider and instantiate it there anymore.

packages/adapters-library/src/core/decorators/cacheToDb.ts
bergarces, line 1: Since we don't have to pass any params to `@CacheToDb()`, I have removed the extra closure. Now the decorator is just `@CacheToDb`.

packages/adapters-library/src/core/utils/erc20Movements.ts
bergarces, line 1: Now that all metadata is inside the protocol token object, it leaks extra fields when using the spread operator, so we need to be more specific.

packages/adapters-library/src/adapters/convex/tests/snapshots/ethereum.withdrawals.2.json
```diff
@@ -115,6 +118,27 @@
...
+          "removed": false,
+          "topics": [
+            "0xe2403640ba68fed3a2f88b7557551d1993f84b99bb10ff833f0cf8db0c5e0486",
+            "0x0000000000000000000000008654a995426e775f3ef023cd6e1b5681e774ffa1"
+          ],
+          "transactionHash": "0x9692d3ab0181a4c0017666e2ca49e9d10fee6b7eef3d05de3460caae4072d1fc",
+          "transactionIndex": "0xb8"
+        }
+      ]
+    },
```
jpsains, line 138: just to make sure this extra log in response is expected?

jpsains (approved):

PR #331, opened 2024-09-30
feat: journal

jpsains:


packages/adapters-library/src/SQLiteMetadataProvider.ts
```diff
@@ -48,6 +48,8 @@ export class SQLiteMetadataProvider implements IMetadataProvider {
 
   constructor(filename: string | Buffer, options: Database.Options) {
     this.database = new Database(filename, options)
+
+    this.database.pragma('journal_mode = DELETE')
```
jpsains, line 52: set back to the default setting must be this setting for lambdas

PR #333, opened 2024-10-03
Error-filter

bergarces:

  - [ ] positions
  - [ ] profits
  - [ ] prices (unwrap)
  - [ ] deposits
  - [ ] withdrawals
  - [ ] repays
  - [ ] borrows
  - [ ] tvl

### DIFF
--- a/packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
+++ b/packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
@@ -27,7 +27,10 @@ import { Erc20Metadata } from '../../types/erc20Metadata'
 import { AdaptersController } from '../adaptersController'
 import { Chain } from '../constants/chains'
 import { IMetadataBuilder } from '../decorators/cacheToFile'
-import { NotImplementedError } from '../errors/errors'
+import {
+  NotImplementedError,
+  ProtocolTokenFilterRequiredError,
+} from '../errors/errors'
 import { CustomJsonRpcProvider } from '../provider/CustomJsonRpcProvider'
 import { filterMapAsync } from '../utils/filters'
 import { getTokenMetadata } from '../utils/getTokenMetadata'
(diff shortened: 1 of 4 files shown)


packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
```diff
@@ -634,13 +635,7 @@ export abstract class UniswapV2PoolForkAdapter implements IProtocolAdapter {
       factoryAddress,
       this.provider,
     )
-    const allPairsLength = 1 ?? Number(await factoryContract.allPairsLength())
```
bergarces, line 637: Right side is never used, so it's left as 1.

packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
```diff
@@ -634,13 +635,7 @@ export abstract class UniswapV2PoolForkAdapter implements IProtocolAdapter {
...
-
-    if (allPairsLength > this.MAX_FACTORY_POOL_COUNT) {
```
bergarces, line 639: Next statement already picks a minimum between `MAX_FACTORY_POOL_COUNT` and `allPairsLength`, so this is not necessary.

packages/adapters-library/src/core/errors/errors.ts
```diff
@@ -143,3 +143,11 @@ export class MulticallError extends BaseError {
...
   }
 }
+
+export class ProtocolTokenFilterRequiredError extends BaseError {
+  constructor(
+    message = 'Too many tokens to fetch, protocolTokenAddresses filter must be provided',
+  ) {
+    super(message)
+  }
+}
```
bergarces, line 153: New error so we can filter it out in the `detect-errors` test.

jpsains (approved):

packages/adapters-library/src/tests/detect-errors.test.ts
```diff
@@ -47,6 +47,7 @@ function filterErrors(
         'NotApplicableError',
         'NotImplementedError',
         'NotSupportedError',
+        'ProtocolTokenFilterRequiredError',
```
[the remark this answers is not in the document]
bergarces: I was sure when BSC and Fantom were constantly erroring there.

There isn't an immediate need for this anymore, but I would have thought detect-errors is for unexpected error. This is expected.

PR #337, opened 2024-10-08
Sushiswap-pools

bergarces:
- Restores token0 and token1 additional metadata that was wrongly deleted in the db migration PR.
- Recreated DB entries for all UniV2 adapters with up to date and correct additional metadata.
- Fixes issue with factory pool extraction for UniV2 forks.
- Fixes a couple of logs so that they display better information when building metadata.


packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
```diff
@@ -680,8 +674,6 @@ export abstract class UniswapV2PoolForkAdapter implements IProtocolAdapter {
               { ...token0, address: getAddress(token0.address) },
               { ...token1, address: getAddress(token1.address) },
             ],
-            token0: getAddress(token0.address),
-            token1: getAddress(token1.address),
```
bergarces, line 684: Not part of the metadata. This was being leaked.

packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
```diff
@@ -624,13 +624,7 @@ export abstract class UniswapV2PoolForkAdapter implements IProtocolAdapter {
       factoryAddress,
       this.provider,
     )
-    const allPairsLength = 1 ?? Number(await factoryContract.allPairsLength())
```
bergarces, line 627: It was always being set to 1.

jpsains (approved):

packages/adapters-library/src/core/adapters/UniswapV2PoolForkAdapter.ts
bergarces: Wrong. This needs to be reverted before merging.

SQL queries do not guarantee the order, even though in practical terms they will be in order. We should not rely on that.

packages/adapters-library/src/adapters/pancakeswap-v2/products/pool/pancakeswapV2PoolAdapter.ts
```diff
@@ -36,10 +36,8 @@ export class PancakeswapV2PoolAdapter extends UniswapV2PoolForkAdapter {
   > {
     return {
       [Chain.Ethereum]: {
-        type: 'graphql',
+        type: 'factory',
         factoryAddress: '0x1097053Fd2ea711dad45caCcc45EfF7548fCB362',
-        subgraphUrl:
-          'https://api.thegraph.com/subgraphs/name/pancakeswap/exhange-eth',
```
bergarces, line 42: It does not exists anymore.

PR #342, opened 2024-10-11
feat: add the basic apy calculator

xavier-brochard:
Adds an interface `ApyCalculator` and a first implementation of it, `BalanceOfApyCalculator` that can calculate APYs for some types of protocols using only data already available in the adapters. 

The unit tests back the calculated APY by comparing it against DefiLlama, over a small sample of pools from Lido, Maker, and AaveV3. 

It's not yet plugged to anything.


packages/adapters-library/src/core/apy-calculators/BalanceOfApyCalculator.ts
```diff
@@ -0,0 +1,122 @@
...
+  // How many periods there is in a year
+  public static readonly FREQUENCY = 365
+
+  public async getApy({
+    userAddress,
+    blockNumber,
+    protocolTokenAddress,
+    chainId,
+    adapter,
+  }: EvmApyArgs): Promise<ApyCalculation> {
```
jpsains, line 33: I noticed you are passing the adapter here.

Incase you didnt know the AdapterController has a method where you can pass just the protocol token address and chainId and itll return the matching adapter.

```
  async fetchTokenAdapter(
    chainId: Chain,
    tokenAddress: string,
  ): Promise<IProtocolAdapter | undefined> {
    // Deferred promise so that only the first execution path does the work
    if (!this.protocolTokens) {
      this.protocolTokens = this.buildProtocolTokens()
    }

    const protocolTokens = await this.protocolTokens

    return protocolTokens.get(chainId)?.get(tokenAddress)
  }

```

packages/adapters-library/src/core/apy-calculators/BalanceOfApyCalculator.ts
```diff
@@ -0,0 +1,122 @@
...
+        userAddress,
+        blockNumber: blocknumberEnd,
+        protocolTokenAddresses: [protocolTokenAddress],
+      }),
+    ])
+
+    const balanceStartWei =
+      positionsStart?.[0]?.tokens?.[0]?.tokens?.[0]?.balanceRaw
+    const balanceEndWei =
+      positionsEnd?.[0]?.tokens?.[0]?.tokens?.[0]?.balanceRaw
```
jpsains, line 56: If the underlying token length is more than 1 I think you could log/error/return _unsupported pool_

```
 if(positionsStart?.[0]?.tokens?.[0]?.tokens.length > 1) {exit...}
```

jpsains: This is an optimisation for another day but we should expose a way for you to be able to check if the protocol token has more than one underlying before doing the getPositions request

jpsains: e.g. we could add on here get pool metadata where you can check the number of underlying tokens

```
export type IMetadataProvider = {
  getMetadata: (input: {
    protocolId: Protocol
    productId: string
  }) => Promise<ProtocolToken[]>

  getPoolCount: (protocolId: Protocol, productId: string) => Promise<number>
}
```

jpsains:
Great work on this 

I was thinking you can plug in the APY and APR at this [location](https://github.com/consensys-vertical-apps/mmi-defi-adapters/blob/main/packages/adapters-library/src/core/getProfits.ts#L317) 

This way we are reusing the positions request that are made with the profit requests. Also then the dashboard team wont need to do an additional request.

On your BalanceOfApyCalculator add a method like the one below (using the logic you've already written)

```
getAPY(startPosition,endPosition)
```

what do you think?

As a result the APY APR will be included in the profits response and also no additional requests are needed. 

You can test the result with 
```
npm run profits {userAddress}
```

One additional consideration: profits can be called for 1d, 7d, or 30d, meaning the t0 to t1 can vary, but I believe your APY logic should support these cases—scaling the interest from 1 day, 7 days, or 30 days to an annualized APY.

Let me know your thoughts!

jpsains (approved):

PR #347, opened 2024-10-16
feat: aave v3 reward adapter

jpsains:
* Aave protocol reward adapter
* Separate adapter created - Note the rewards are generated and linked to an aToken position however when the reward is claimed there is no event linking the amount claimed with the aToken position. IMO this solution is good enough but not ideal due to the missing event.
* Also debank does not link the reward to a aToken

image


packages/adapters-library/src/adapters/aave-v2/common/aaveBasePoolAdapter.ts
```diff
@@ -192,22 +193,20 @@ export abstract class AaveBasePoolAdapter implements IProtocolAdapter {
...
 
       const [protocolToken, underlyingToken] = await Promise.all([
         protocolTokenPromise,
         underlyingTokenPromise,
       ])
 
+      if (protocolToken.address === ZERO_ADDRESS) {
+        return
+      }
+
```
jpsains, line 209: Zero address was bing added as a pool address

context: seems aave removed stable interest debt token for eth

packages/adapters-library/src/scripts/buildMetadataDb.ts
```diff
@@ -155,6 +156,31 @@ async function buildAdapterMetadata(adapter: IProtocolAdapter) {
...
+      chalk.green('\n * Please ensure that address pool address is correct'),
+    )
+    console.error(
+      chalk.green(
+        '\n * Please checksum your addresses inside the buildMetadata() method.',
+      ),
+    )
+    return
+  }
+
```
jpsains, line 183: ensure we dont try to save a pool with the zero address

packages/adapters-library/src/scripts/buildMetadataDb.ts
```diff
@@ -155,6 +156,31 @@ async function buildAdapterMetadata(adapter: IProtocolAdapter) {
     return
   }
 
+  const invalidPoolAddress = metadataDetails.find((pool) => {
+    if (pool.address === ZERO_ADDRESS) {
+      return true
+    }
+  })
```
bergarces, line 163:
```suggestion
  const invalidPoolAddress = metadataDetails.find((pool) => pool.address === ZERO_ADDRESS)
```

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
```diff
@@ -0,0 +1,391 @@
...
+    return [
+      { ...this.INCENTIVES_CONTRACT_DETAILS, underlyingTokens: rewardTokens },
+    ]
+  }
+
+  private detectPositionEventSignature(
+    eventSignature = 'Transfer(address,address,uint256)',
+  ): string {
+    return ethers.id(eventSignature)
+  }
```
jpsains, line 161: @bergarces please internalise this - I think we need to make it part of the IProtocolAdapter so that we can make all adapters eth-event driven

packages/adapters-library/src/scripts/buildMetadataDb.ts
```diff
@@ -155,6 +156,31 @@ async function buildAdapterMetadata(adapter: IProtocolAdapter) {
...
+
+  if (invalidPoolAddress) {
+    console.error(chalk.yellow(ZERO_ADDRESS))
```
bergarces, line 166:
```suggestion
    console.error(chalk.yellow(invalidPoolAddress.address))
```
A bit unsafe not to use the address of the pool returned otherwise, just in case we change the condition.

packages/adapters-library/src/scripts/buildMetadataDb.ts
bergarces: We didn't we add the ZERO_ADDRESS to the method that checks invalid addresses?

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
```diff
@@ -0,0 +1,391 @@
...
+
+    const filteredLogs = logs
+      .filter((log) => aTokenAddresses.includes(log.address))
+      .map((log) => log.address)
+
+    return [...new Set(filteredLogs)]
+  }
+
+  async getPositions(input: GetPositionsInput): Promise<ProtocolPosition[]> {
+    const { userAddress, blockNumber, protocolTokenAddresses } = input
```
bergarces, line 188: I don't think we are using the `protocolTokenAddresses` filter.

jpsains: ah i should try to use it i think e.g. return if not this protocolTokenAddress

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
```diff
@@ -0,0 +1,400 @@
...
+  private INCENTIVES_CONTRACT_ADDRESSES = {
+    [Chain.Arbitrum]: getAddress('0x929EC64c34a17401F460460D4B9390518E5B473e'),
+    [Chain.Ethereum]: getAddress('0x8164Cc65827dcFe994AB23944CBC90e0aa80bFcb'),
+    [Chain.Fantom]: getAddress('0x929EC64c34a17401F460460D4B9390518E5B473e'),
+    [Chain.Avalanche]: getAddress('0x929EC64c34a17401F460460D4B9390518E5B473e'),
+    [Chain.Bsc]: getAddress('0xC206C2764A9dBF27d599613b8F9A63ACd1160ab4'),
+    [Chain.Base]: getAddress('0xf9cc4F0D883F1a1eb2c253bdb46c254Ca51E1F44'),
+    [Chain.Polygon]: getAddress('0x929EC64c34a17401F460460D4B9390518E5B473e'),
+    [Chain.Optimism]: getAddress('0x929EC64c34a17401F460460D4B9390518E5B473e'),
+    [Chain.Linea]: '', // not deployed on linea
```
xavier-brochard, line 65: Should we instead just omit the entry? Using empty string will break on line 87 no? 👇 
```
    this.incentivesContract = IncentivesContract__factory.connect(
      this.INCENTIVES_CONTRACT_ADDRESSES[this.chainId],
      this.provider,
    )
```

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
```diff
@@ -0,0 +1,400 @@
...
+    // If no claimable underlying tokens are found, return empty
+    if (underlyingTokens.length === 0) {
+      return []
+    }
+
+    // Return the final ProtocolPosition object
+    return [
+      {
+        type: TokenType.Protocol,
+        balanceRaw: 1n, // choose 1 here as a zero value may cause the position to be ignored on UIs our adapters currently expecting a protocol token but on contract positions there is no token
```
xavier-brochard, line 232: Oof

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
```diff
@@ -0,0 +1,400 @@
...
+              type: TokenType.Underlying,
+              blockNumber,
+            },
+          ],
+          blockNumber,
+        }
+      }),
+    )
+
+    return movements.filter((movement) => movement !== undefined)
```
xavier-brochard, line 368:
```suggestion
    return movements.filter(Boolean)
```

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
bergarces: Do you mean to use it when we cannot take advantage of the one we run automatically outside the adapter?

An idea here would be to create a new public method that returns open positions based on events and use the questionare answers to see if we can populate it with a helper (which is ok for most tokens). Then we can always check if it's implemented and use it outside the adapter, instead of relying on the developer to do the filtering.

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
bergarces: I think this is a bad idea too.

If the client is ignoring it because it shows 0, it's the client's problem.

bergarces: We have an open ticket to deal with this sort of scenarios, maybe the answer is for contract positions to not have a balance at all, just underlyings.

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
bergarces: We only support this adapter on:
- Ethereum
- Polygon
- Avalanche
- Base
- Arbitrum
- Optimism
- Fantom (it's commented, but it's there)

We should remove this line for Linea.

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
jpsains: I mean instead of running it outside the adapter we move it to inside the adapters. 

Many will be Transfer(address, address, amount)

packages/adapters-library/src/adapters/aave-v3/products/rewards/aaveV3RewardsAdapter.ts
jpsains: Hey, 

Did you guys know the balance means different things on a standard erc20 token vs a standard erc721

Reminder:
```
ERC-20: Represents a countable, divisible balance of identical tokens.
ERC-721: Represents ownership of distinct, indivisible assets where each token is unique and tracked by its own ID.
```

@bergarces maybe it does make sense to have a balance but for contract and nft positions it represents the number of distinct positions held.

jpsains: Thoughts?

So 1n should be underlying.length and decimals 1

jpsains: maybe for contract and nft positions we should drop the decimals but yeah not sure atm

PR #351, opened 2024-10-17
fix dev ui with express backend

bergarces:


packages/dev-backend/index.js
bergarces, line 1: I wasn't going to bother with TypeScript dependencies for this setup, it's just an express wrapper over the our DefiProvider.

.npmrc
```diff
@@ -1 +1,2 @@
+registry=https://registry.npmjs.org/
```
bergarces, line 1: This stops jfrog being used as the registry.

packages/dev-ui/src/chainIdToChainNameMap.ts
bergarces, line 1: Unfortunately we cannot use the one from the library without going through a lot of trouble. This is a good compromise given we don't add chains very often.

jpsains (approved):

PR #352, opened 2024-10-17
feat: price improvement

jpsains:


packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -1,41 +1,78 @@
...
+  tokenAddresses: string
+  vsCurrency?: string
+  chainId: number
+}
+
+export type IPricesInput = {
+  blockNumber?: number
+  tokenMetadata: Erc20Metadata
+}
+export interface IPricesAdapter {
```
bergarces, line 54: Very glad you went for a different interface. We shouldn't be afraid of doing it when it feels too forced and it has drawbacks (such as not having the token metadata).

packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -1,41 +1,78 @@
...
+  totalVolume?: number
+}
+
+export interface GetSpotPriceByAddressInput {
+  tokenAddresses: string
+  vsCurrency?: string
+  chainId: number
+}
+
+export type IPricesInput = {
```
bergarces, line 50: nit: No `I` for inputs.

packages/adapters-library/src/unwrapCache.ts
```diff
@@ -47,6 +57,34 @@ export class UnwrapCache implements IUnwrapCache {
 
     await this.unwrapCacheProvider.setToDb(key, value)
 
+    return value
+  }
+  async fetchPriceWithCache(
```
bergarces, line 62: It might be possible to make an abstraction in which instead of the adapter and the input, we pass a runner function, chainId, token address and block number, but I don't really think it's worth it at this point.

Maybe if we revisit things in the future.

bergarces (dismissed):

bergarces:
Fix tests before merging.

PR #354, opened 2024-10-21
feat: filter fix

jpsains:
* Added validation to protect us from being spammed, now productid, protocolid, chainid and one protocol token address must be supplied.
* Rewriting interfaces to accept strings instead of array strings not worth the investment since this may be temporary and also feature not used anywhere (it would mean rewriting scripts and perhaps touching every tvl implementation on each adapter )

* TVL was spamming the price adapter with many requests despite one protocol token address being supplied
* That meant protocol token address filter is not correctly implemented caught issues with about 5 adapters not correctly filtering on protocol token addresses


packages/adapters-library/src/defiProvider.ts
```diff
@@ -800,6 +803,36 @@ export class DefiProvider {
...
+        `No adapter found for protocol address: ${protocolAddress} on chain: ${chainId}`,
+      )
+    }
+
+    if (adapter.protocolId !== protocolId || adapter.productId !== productId) {
+      throw new TvlValidationError(
+        `Protocol address ${protocolAddress} adapter does not match the provided protocolId: ${protocolId} and productId ${productId}`,
+      )
+    }
+
```
jpsains, line 835: Previous implementation was spamming price adapter.

Now instead first we find the correct adapter and run tvl only for that adapter.

packages/adapters-library/src/defiProvider.ts
```diff
@@ -808,6 +841,12 @@ export class DefiProvider {
         blockNumber,
       })
 
+      if (tokens.length > 1) {
+        throw new TvlValidationError(
+          'Total value locked should return maximum one protocol token, adapter incorrectly implemented',
+        )
+      }
+
```
jpsains, line 849: If protocol token length is more than one that mean the adapter has not applied the protocol token address filter

bergarces (dismissed):

PR #363, opened 2024-10-28
Mmi 5676 slack channel qa investigate missing products and pools

jpsains:


packages/adapters-library/src/core/adaptersController.ts
```diff
@@ -299,6 +301,22 @@ export class AdaptersController {
...
+            .getProtocolTokens()
+            .catch(() => undefined)
+
+          if (protocolTokens) {
+            if (!product.protocolTokens) {
+              product.protocolTokens = {}
+            }
+
+            product.protocolTokens[chainId] = protocolTokens
+          }
```
jpsains, line 319: Now can see full protocol tokens with underlying when running support with flag -pt

packages/adapters-library/src/scripts/checksumAddress.ts
bergarces, line 1: This should be under `core/decorators`.

packages/adapters-library/src/defiProvider.ts
```diff
@@ -873,6 +883,7 @@ export class DefiProvider {
   async getSupport(input?: {
     filterChainIds?: Chain[] | undefined
     filterProtocolIds?: Protocol[] | undefined
+    includeProtocolTokens?: boolean
```
bergarces, line 886: With this new parameter added. Shouldn't it also use the new decorator?

packages/adapters-library/src/scripts/checksumAddress.ts
jpsains: Thanks, was in process of moving to utils but decorators much better

packages/adapters-library/src/defiProvider.ts
jpsains: nope......this is a boolean to include protocol tokens it is not a address input

PR #364, opened 2024-10-28
feat: pendle

jpsains:


packages/adapters-library/src/adapters/aave-v2/common/aaveBasePoolAdapter.ts
```diff
@@ -176,18 +176,6 @@ export abstract class AaveBasePoolAdapter implements IProtocolAdapter {
...
-        await protocolDataProviderContract.getReserveConfigurationData(
-          tokenAddress,
-        )
-
-      if (
-        !reserveConfigurationData.isActive ||
-        reserveConfigurationData.isFrozen
-      ) {
-        return
-      }
```
bergarces, line 189: So... we want to show inactive and frozen pools?

I don't remember why we added this in the first place. Is there any example that wasn't showing before?

packages/adapters-library/src/adapters/pendle/products/yield-token/pendleYieldTokenAdapter.ts
```diff
@@ -103,14 +130,26 @@ export class PendleYieldTokenAdapter implements IProtocolAdapter {
...
+    const tokens = await filterMapAsync(
+      await this.getProtocolTokens(),
+      async (protocolToken) => {
+        if (await this.isExpiredAtBlock(protocolToken.expiry)) {
+          return undefined
+        }
+
+        return protocolToken
+      },
+    )
```
bergarces, line 148: At least add a TODO here, because when we can update the metadata db dynamically we 1000000% don't want to do this, as it's just delaying fetching positions whilst waiting for block info.

bergarces (approved):

packages/adapters-library/src/adapters/pendle/products/yield-token/pendleYieldTokenAdapter.ts
jpsains: Hey,

I see - you are saying we run metadata daily then save isExpired to true in db, interesting 

You found a good use case to run the metadata daily.

PR #365, opened 2024-10-29
Convex improvements

bergarces:
Fixed issues:
- Bug that duplicated underlying, rewards and extra rewards when fetching the data from the DB
- Wrong name for wsteth token in the db
- Duplicate integration tests for checksum addresses fixed
- Remove async from `getProtocolTokenByAddress`

Convex
- Reimplemented pool and staking adapters using `IProtocolAdapter`.
- Better handling of rewards for staking adapter.
- Added event filtering when getting positions.


packages/adapters-library/src/scripts/helpers.ts
```diff
@@ -42,15 +42,15 @@ export class Helpers {
     public readonly allJsonRpcProviders: Record<Chain, CustomJsonRpcProvider>,
   ) {}
 
-  async getProtocolTokenByAddress<
```
bergarces, line 45: Any issue removing the `async` here? It shouldn't be asynchronous and it's adding overhead.

packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -130,7 +131,7 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
       const data: SpotPrice = await response.json()
       return data
     } catch (error) {
-      console.error('Failed to fetch spot price:', error)
```
bergarces, line 133: Using `logger` on everything that happens at runtime.

packages/adapters-library/src/adapters/integration.test.ts
bergarces, line 1: I noticed that the checksum check was running for every product multiple times. This makes it run only for the current product and avoids repetition.

packages/adapters-library/src/SQLiteMetadataProvider.ts
bergarces, line 1: Bug discovered here.

When the pool has more than one underlying, or reward, or extra reward, multiple rows are returned, so we need to make sure we process each item once.

In the case of convex, with 1 underlying, 2 rewards and 1 extra reward. What was being returned were 2 underlyings (duplicates), 2 rewards and 2 extra rewards (duplicates). It was somehow working because we were always doing `underlying[0]` in combination with `SimplePoolAdapter`, but now that I've rewritten it to use the helpers it unveiled the issue.

bergarces, line 1: Bug discovered here.

When the pool has more than one underlying, or reward, or extra reward, multiple rows are returned, so we need to make sure we process each item once.

In the case of convex, with 1 underlying, 2 rewards and 1 extra reward. What was being returned were 2 underlyings (duplicates), 2 rewards and 2 extra rewards (duplicates). It was somehow working because we were always doing underlying[0] in combination with SimplePoolAdapter, but now that I've rewritten it to use the helpers it unveiled the issue.

packages/adapters-library/src/adapters/integration.test.ts
bergarces, line 1: I noticed that the checksum check was running for every product multiple times. This makes it run only for the current product and avoids repetition.

packages/adapters-library/src/scripts/featureCommands.ts
```diff
@@ -139,49 +140,60 @@ function addressCommand(
       '-t, --protocol-tokens <protocol-tokens>',
       'comma-separated protocol token address filter (e.g. 0x030..., 0x393.., 0x332...)',
     )
+    .option(
+      '-pd, --product-ids <product-ids>',
+      'comma-separated product id filter (e.g. reward, a-token, staking)',
+    )
```
bergarces, line 146: Added option to select productId

packages/adapters-library/src/adapters/lido/products/wst-eth/tests/snapshots/ethereum.deposits.3.json
```diff
@@ -17,14 +17,14 @@
         "transactionHash": "0x779025a8260db4601f74506e6af804edcb7344b606f9bad07c59e3706c82f144",
         "protocolToken": {
           "address": "0x7f39C581F595B53c5cb19bD0b3f8dA6c935E2Ca0",
-          "name": "wstETH",
+          "name": "Wrapped liquid staked Ether 2.0",
```
bergarces, line 20: This is the correct name for that token. There's some adapter in our DB that keeps updating it with the wrong name.

PR #396, opened 2025-02-13
feat: latest block inexer

jpsains:


packages/workers/src/db-queries.ts
bergarces, line 1: Great changes here. They'll make the functions more reusable.

In the PR I am preparing to enable BSC I'll move the history cache table definitions to the `build-historic-cache.ts` file as well.

I'll try to keep this file as generic as possible.

packages/workers/src/block-indexer.ts
```diff
@@ -0,0 +1,173 @@
...
+    this._processBlockFn = processBlockFn
+    this._getStartBlockNumber = getStartBlockNumberFn
+    this._onError = onError
+
+    this._provider = provider
+    this._chainId = chainId
+    this._chainName = ChainIdToChainNameMap[chainId]
+  }
+
+  private async waitForNewBlockToProcess(currentBlock: number): Promise<void> {
```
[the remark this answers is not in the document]
bergarces: From genesis to the latest block. I have been able to index every single transfer event for every chain and every pool we support (some chains up to 14K pools) in less than 12 hours.

packages/workers/src/block-indexer.ts
```diff
@@ -0,0 +1,173 @@
...
+    onError: (latestSafeProcessedBlock: number) => Promise<void>
+  }) {
+    this._chainName = ChainName[chainId]
+    this._processBlockFn = processBlockFn
+    this._getStartBlockNumber = getStartBlockNumberFn
+    this._onError = onError
+
+    this._provider = provider
+    this._chainId = chainId
+    this._chainName = ChainIdToChainNameMap[chainId]
```
bergarces, line 42: This is already being setup on line 35.

When exporting `ChainName` from `@metamask-institutional/defi-adapters` directly, it is already exporting the correct one (`ChainIdToChainNameMap`).

packages/workers/src/block-indexer.ts
bergarces, line 1: Can we call this file `block-runner.ts` to match the class name?

packages/workers/src/block-indexer.ts
```diff
@@ -0,0 +1,173 @@
...
+
+          logger.info(
+            `[${this._chainName}] Max retries exceeded for block ${processingBlockNumber}`,
+          )
+        }
+        await new Promise((res) => setTimeout(res, 5000 * retryCount))
+      }
+    }
+
+    async function logUpdate(chain: Chain, provider: CustomJsonRpcProvider) {
```
bergarces, line 160:
```suggestion
    private async function logUpdate(chain: Chain, provider: CustomJsonRpcProvider) {
```

Should be private probably.

packages/workers/src/block-indexer.ts
jpsains: yep good catch

packages/workers/src/block-indexer.ts
```diff
@@ -0,0 +1,173 @@
...
+          error instanceof Error ? error.stack : String(error),
+        )
+
+        logger.info(
+          `curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x${processingBlockNumber.toString(
+            16,
+          )}", true],"id":1}' -H "Content-Type: application/json" ${
+            this._provider._getConnection().url
+          } | jq`,
+        )
```
bergarces, line 143: Is this intended? It's a bit of an odd thing to log.

jpsains: dont need it now - was handy at the start

jpsains: will remove

packages/workers/src/block-indexer.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+
+  async start() {
+    logger.info(`[${this._chainName}] Starting block indexer...`)
+    let processingBlockNumber = await this._getStartBlockNumber()
+    this._latestBlockNumber = await this._provider.getBlockNumber()
+    let retryCount = 0
+
+    while (retryCount <= BlockRunner._MAX_RETRIES) {
```
bergarces, line 74: Why are they a maximum of retries?

We do not want the process to ever stop, even if it keeps failing. We can then rely on errors logs to alert us know if the process hasn't been healthy for a while.

packages/workers/src/build-latest-cache.ts
```diff
@@ -0,0 +1,200 @@
...
+  db,
+}: {
+  provider: CustomJsonRpcProvider
+  blockNumber: number
+  watchContractListLowercase: Set<string>
+  db: DatabaseType
+}): Promise<void> {
+  const receipts = await provider.send('eth_getBlockReceipts', [
+    `0x${ethers.toBeHex(blockNumber).slice(2).replace(/^0+/, '')}`,
+  ])
```
bergarces, line 160: I've just tested `eth_getBlockReceipts` with multiple leading 0s and the result is correct.

```
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "eth_getBlockReceipts",
    "params": ["0x00001324bf8"]
}
```

Why are we removing the leading 0s from the number here?

Also, does `ethers.toBeHex` even add leading 0s?

packages/workers/src/build-latest-cache.ts
```diff
@@ -0,0 +1,200 @@
...
+  const queries: string[] = []
+
+  for (const receipt of receipts?.flat() || []) {
+    for (const log of receipt.logs || []) {
+      // retuned lowercase from provider
+      const contractAddressLowercase = log.address
+      if (watchContractListLowercase.has(contractAddressLowercase.slice(2))) {
+        log.topics
+          .filter((topic: string) =>
+            topic.startsWith('0x000000000000000000000000'),
```
bergarces, line 171: Can you add to the filter that `topic !== '0x0000000000000000000000000000000000000000000000000000000000000000'`? We do not want to store references to the zero address in the DB.

bergarces (approved):

PR #397, opened 2025-02-18
Move scripts to adapters-cli

bergarces:
- Moves scripts from adapters-library to adapters-cli.
- Remove script dependencies from adapters-library
- Prunes docker images so that only production dependencies are kept
- Template snapshots are now in adapters-cli, using vitest as it's friendlier with esm.


.github/workflows/verify-build.yaml
```diff
@@ -152,6 +152,29 @@ jobs:
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Run test
+        run: npm run test -w packages/adapters-cli
+        env:
+          DEFI_ADAPTERS_LOG_LEVEL: warn
```
bergarces, line 176: Move this to its own job, as it's now in a different package and it was difficult to open the logs on the previous job due to the size of this test.

packages/adapters-library/src/tests/detect-errors.test.ts
```diff
@@ -38,6 +38,7 @@ function filterErrors(
         'NotImplementedError',
         'NotSupportedError',
         'ProtocolTokenFilterRequiredError',
+        'NotSupportedUnlimitedGetLogsBlockRange',
```
bergarces, line 41: This stops the detect-errors test from failing.

Dockerfile.workers
```diff
@@ -12,6 +12,9 @@ RUN npm ci
 
 COPY . .
 
-RUN npm run build
+RUN npm run build -w packages/adapters-library
+RUN npm run build -w packages/workers
```
jpsains, line 16: might be worth creating a npm script?

Dockerfile.api
```diff
@@ -5,12 +5,14 @@ WORKDIR /app
...
 COPY packages/adapters-api/package*.json ./packages/adapters-api/
-COPY packages/workers/package*.json ./packages/workers/
 
 RUN npm ci
 
 COPY . .
 
-RUN npm run build
+RUN npm run build -w packages/adapters-library
+RUN npm run build -w packages/adapters-api
```
jpsains, line 14: might be worth creating a npm script?

jpsains (approved):

packages/adapters-library/src/tests/detect-errors.test.ts
jpsains: i dont think it has

jpsains: looks like its failed still

bergarces: It failed for something else. All of it BSC-related...

PR #407, opened 2025-03-14
Group positions metadata

bergarces:


packages/adapters-api/src/index.ts
```diff
@@ -49,7 +51,9 @@ app.get('/support', async (context) => {
     const input = context.req.query()
     const parsedInput = GetSupportSchema.parse(input)
 
-    return context.json(await defiProvider.getSupport(parsedInput))
+    return context.json({
+      data: await defiProvider.getSupport(parsedInput),
+    })
```
bergarces, line 56: Doing this to make it compatible with the current API from the lambdas. It will also enable of us to return timestamp of the data if we need to.

PR #408, opened 2025-03-19
postgres update for worker

bergarces:
- Workers postgres refactor
- Uniswap-V3 event added and metadata updated
- `ChainName` is now the only way of getting chain names
- It is possible to disable chains by passing an empty string as the rpc url


packages/adapters-cli/src/commands/build-cache-commands.ts
bergarces, line 1: It gets a bit complicated to run this from the CLI with the postgres setup, makes more sense to run it with `docker compose up`

packages/adapters-library/src/config.ts
```diff
@@ -215,10 +218,11 @@ function parseNumberEnv(env: string | undefined, defaultValue: number): number {
 }
 
 function parseStringEnv(env: string | undefined, defaultValue: string): string {
-  // Uses defaultValue if empty string is passed
-  return env || defaultValue
+  // Uses defaultValue only if env is undefined
+  return env ?? defaultValue
```
bergarces, line 222: This is only used for the provider names. Doing this, if we don't pass any provider URL, it'll use the default as it'll be undefined, but if we pass an empty string we can then disable that provider without the need for another environment variable (handy for BSC and Fantom without having to comment code).

packages/adapters-library/src/core/adaptersController.ts
bergarces, line 1: Some changes to this file in order to allow a partial list of providers, to accommodate disabling chains. This means we need to check if the provider is set at the beginning of some flows.

packages/adapters-library/src/core/helpers.ts
```diff
@@ -44,10 +44,6 @@ export class Helpers {
     public readonly provider: CustomJsonRpcProvider,
     public readonly chainId: EvmChain,
     public readonly metadataProvider: IMetadataProvider,
-    public readonly allJsonRpcProviders: Record<
-      EvmChain,
-      CustomJsonRpcProvider
-    >,
```
bergarces, line 50: This wasn't being used at all.

databases/op.db
jpsains, line 1: you dont have a db update for binance

jpsains (dismissed):

PR #410, opened 2025-03-24
feat: add icon urls

jpsains:


packages/adapters-library/src/core/adaptersController.ts
```diff
@@ -323,33 +323,37 @@ export class AdaptersController {
...
+              if (!product.protocolTokenAddresses) {
+                product.protocolTokenAddresses = {}
+              }
+  
+              product.protocolTokenAddresses[chainId] = protocolTokenAddresses
+            }
+  
+            if (!includeProtocolTokens) {
+              continue
+            }
```
bergarces, line 341: This if is redundant now.

PR #412, opened 2025-03-26
feat: liquid staking adapter

jpsains:


packages/adapters-library/src/adapters/metamask/products/pooled-staking/tests/snapshots/ethereum.positions.json
```diff
@@ -0,0 +1,73 @@
...
+          ],
+          "balance": 0.048806725151497525
+        }
+      ]
+    }
+  ],
+  "rpcResponses": {
+    "22fafb1e961ef43f59f28cbba6538494": {
+      "result": []
+    },
```
jpsains, line 59: @bergarces this is the transfer log request

bergarces: 👍 
Even if they take more space, I think we ought to seriously consider setting `DEFI_ADAPTERS_SAVE_INTERCEPTED_REQUESTS` to true so that it also saves the rpc request, not just the hash and the response.

nitro-devnode
bergarces, line 1: Can we get rid of this?

packages/adapters-library/src/adapters/metamask/products/pooled-staking/metamaskPooledStakingAdapter.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    if (protocolTokens.length !== 1) {
+      throw new Error('Only one protocol token is supported')
+    }
+
+    const { underlyingTokens, ...protocolToken } = protocolTokens[0]!
+
+    if (underlyingTokens.length !== 1) {
+      throw new Error('Only one underlying token is supported')
+    }
```
bergarces, line 134: I'm not a big fan of this unreachable throws. If we know what the shape of the adapter metadata is, we can just go with `!`.

Alternative, doing an `assert(condition, errorMessage)` would be cleaner and won't decrease code coverage.

bergarces (dismissed):

bergarces:
Ive solved the conflict with integration tests and removed that nitro-dev folder.

packages/adapters-library/src/adapters/metamask/products/pooled-staking/metamaskPooledStakingAdapter.ts
```diff
@@ -0,0 +1,144 @@
...
+  protocolId: Protocol
+  chainId: Chain
+  helpers: Helpers
+
+  adapterSettings: AdapterSettings = {
+    includeInUnwrap: true,
+    userEvent: {
+      topic0:
+        '0x861a4138e41fb21c121a7dbb1053df465c837fc77380cc7226189a662281be2c',
+      userAddressIndex: 2,
```
bergarces, line 41: I changed this to index 2.

Index 1 is the caller, that is not necessarily who gets the position, which is the receiver in index 2.

PR #413, opened 2025-03-28
Sonarcloud and new tests

bergarces:


packages/adapters-api/package.json
bergarces, line 1: Done so that we can easily ignore this file for coverage with a single rule.

packages/adapters-api/src/build-api.ts
bergarces, line 1: Moved api-building outside main.ts so it can be tested.

PR #417, opened 2025-04-04
dev changes

bergarces:


Dockerfile.workers
bergarces, line 1: No more pm2.

docker-compose.yaml
```diff
@@ -18,6 +18,7 @@ services:
       - DEFI_ADAPTERS_USE_POSITIONS_CACHE=true
       - USE_ADAPTERS_WITH_USER_EVENT_ONLY=true
       - CACHE_DATABASE_URL=postgresql://postgres:postgres@postgres:5432/defi-adapters
+      - CACHE_DATABASE_DISABLE_SSL=true
```
bergarces, line 21: SSL needs to be enabled for AWS postgres. It is better to have that as a default an just disable it for local use.

packages/adapters-api/src/db-service.ts
```diff
@@ -0,0 +1,246 @@
...
+      (acc, curr) => {
+        acc[curr.chainId] = curr.stats
+        return acc
+      },
+      {} as Partial<Record<EvmChain, T>>,
+    )
+  }
+}
+
+export class NoDbService implements DbService {
```
bergarces, line 224: This covers when we deploy the api locally with no postgres/cache.

packages/workers/src/main.ts
```diff
@@ -1,24 +1,17 @@
...
-  process.exit(1)
-}
-
-await runner(process.env.DATABASE_URL, chainId)
+await Promise.all(
+  Object.values(EvmChain).map(async (chainId) => {
+    const childLogger = logger.child({ chainId })
+    await runner(process.env.CACHE_DATABASE_URL!, chainId, childLogger)
+  }),
+)
```
bergarces, line 17: Instead of pm2, it just creates one promise for each chain (that should never finish).

The logger with the chainId needs to be created and passed down as a prop.

packages/workers/src/block-runner.ts
```diff
@@ -71,60 +78,43 @@ export class BlockRunner {
     return true
   }
 
-  async start(startBlock: number) {
```
bergarces, line 74: When doing the logging I was getting a bit confused with all the names `startBlock`, `processingBlockNumber` and `latestBlockNumber`.

Since `startBlock` is used to initialise `processingBlockNumber`, I made it so that is done at constructor time, which has the added advantage that `processingBlockNumber` is now part of the class state and it is easier to access it for logging the value.

jpsains (approved):

PR #422, opened 2025-04-16
adapter fixes and new field

bergarces:


.github/workflows/verify-build.yaml
bergarces, line 1: This is just so that it prints what tests passed and failed in the CI, as it does not with just `blob`.

packages/adapters-library/src/adapters/snapshots.test.ts
```diff
@@ -321,3 +326,12 @@ async function loadJsonFile(
...
 }
+
+// This is to avoid having to update every snapshot when a new field is added to the response
+function sanitizeSnapshot(snapshot: AdapterResponse<unknown>[]) {
+  for (const position of snapshot) {
+    if (position.success) {
+      position.protocolDisplayName = ''
+    }
+  }
+}
```
bergarces, line 337: We can save ourselves having to update every single snapshot when we add a field like this.

packages/adapters-library/src/adapters/metamask/products/pooled-staking/tests/snapshots/ethereum.positions.json
```diff
@@ -9,9 +9,10 @@
       "description": "Metamask defi adapter",
       "siteUrl": "https://portfolio.metamask.io/",
       "iconUrl": "https://portfolio.metamask.io/favicon.png",
-      "positionType": "supply",
+      "positionType": "stake",
       "chainId": 1,
       "productId": "pooled-staking",
+      "protocolDisplayName": "MetaMask Staking",
```
bergarces, line 15: This is what it'll return in the response.

bergarces:
I'll apply them to the other repo so that it gets deployed in dev. (I hope they move the CI here soon)

jpsains (approved):

PR #424, opened 2025-04-23
Fix api

bergarces:


packages/adapters-api/src/build-api.ts
```diff
@@ -1,131 +1,237 @@
...
+          UNPROCESSABLE_ENTITY,
+        )
+      }
+    },
+  })
+}
+
+function addMiddleware(app: ReturnType<typeof createApp>) {
   app.use('*', cors())
+  app.use('*', requestId())
```
bergarces, line 48: This is to add a requestId to the request, which helps with the logs.

packages/adapters-api/src/build-api.ts
```diff
@@ -1,131 +1,237 @@
...
+  app.use(
+    '*',
+    pinoLogger({
+      pino: logger,
+    }),
+  )
```
bergarces, line 54: Passes our logger as middleware so it can log response results (validation errors, success, internal errors, etc)

packages/adapters-api/src/build-api.ts
```diff
@@ -1,131 +1,237 @@
...
 
-  app.get('/health', (context) => context.json({ message: 'ok' }))
+  app.onError((err, c) => {
```
bergarces, line 56: This formats the error response and status code if we throw from within the API.

packages/adapters-api/src/build-api.ts
bergarces, line 1: Only adding API spec to positions and support, as the other endpoints are more for debugging purposes.

jpsains (approved):

jpsains (dismissed):

PR #426, opened 2025-04-25
modify workers and db to accept event data

bergarces:


packages/adapters-library/src/core/adapters/CompoundV2BorrowMarketForkAdapter.ts
```diff
@@ -30,7 +30,11 @@ export abstract class CompoundV2BorrowMarketForkAdapter
 
   adapterSettings: AdapterSettings = {
     includeInUnwrap: false,
-    userEvent: false,
+    userEvent: {
+      eventAbi:
+        'event Borrow(address borrower, uint borrowAmount, uint accountBorrows, uint totalBorrows)',
+      userAddressArgument: 'borrower',
+    },
```
bergarces, line 37: This is how we define how to extract user address from an event when it is not indexed in a topic.

jpsains: How would you feel if I used this method with argument that was indexed? Whats our policy?

packages/workers/src/event-parsing.ts
```diff
@@ -0,0 +1,32 @@
...
+) {
+  let userAddress: string | undefined
+
+  if (eventAbi) {
+    const iface = new Interface([eventAbi])
+    const decoded = iface.parseLog(log)
+
+    if (!decoded) {
+      throw new Error('Log does not match event abi')
+    }
```
jpsains, line 16: are we sure we want to throw.......

jpsains (dismissed):

packages/workers/src/event-parsing.ts
bergarces: Yes. This function is not responsible of logging as it doesn't have full context.

If it detects something that it prevents it from getting an address, we throw an error and then the parent can log it (including the error message) and continue with the loop.

PR #432, opened 2025-05-13
better logs and disable chains

bergarces:


packages/workers/src/main.ts
```diff
@@ -9,9 +9,16 @@ if (!process.env.CACHE_DATABASE_URL) {
 
 logger.info('Starting workers')
 
+const defiProvider = new DefiProvider()
```
bergarces, line 12: Declaring it inside the `runner` meant multiple initializations, each of them repeating certain logs.

packages/adapters-library/src/core/provider/ChainProvider.ts
```diff
@@ -40,7 +40,6 @@ export class ChainProvider {
       (providers, chainId) => {
         const chainName = ChainName[chainId]
 
-        // Throw an error if the provider URL is missing for this chain
```
bergarces, line 43: Removed this comment because it shouldn't throw.

packages/adapters-library/src/config.ts
bergarces, line 1: This is temporary, but it ensures that those chains cannot be used for now.

Ideally, we would add the entry to the deployment secrets and leave it blank, but since we don't know how long it would take for someone to do it in prod, we can take control of it here for now.

docker-compose.yaml
```diff
@@ -8,9 +8,9 @@ services:
     environment:
       - DEFI_ADAPTERS_PROVIDER_ETHEREUM
       - DEFI_ADAPTERS_PROVIDER_OPTIMISM
-      - DEFI_ADAPTERS_PROVIDER_BSC=
+      - DEFI_ADAPTERS_PROVIDER_BSC
       - DEFI_ADAPTERS_PROVIDER_POLYGON
-      - DEFI_ADAPTERS_PROVIDER_FANTOM=
+      - DEFI_ADAPTERS_PROVIDER_FANTOM
```
bergarces, line 13: No more cheating, local env needs to be as close as the deployed ones as possible.

packages/workers/src/runner.ts
```diff
@@ -26,19 +31,9 @@ export async function runner(dbUrl: string, chainId: EvmChain, logger: Logger) {
     logger,
   })
 
-  // TODO: Remove this once we have support for BSC and Fantom
-  if (chainId === Chain.Bsc || chainId === Chain.Fantom) {
```
bergarces, line 30: Removed this from here. It'll use the library defined providers to determine if the chain is enabled or not, the same way the API already does it.

packages/workers/src/runner.ts
```diff
@@ -26,19 +31,9 @@ export async function runner(dbUrl: string, chainId: EvmChain, logger: Logger) {
...
-
-    return
-  }
-
-  const defiProvider = new DefiProvider()
-
   const providerUrl =
     defiProvider.chainProvider.providers[chainId]?._getConnection().url
 
-  // TODO: Should exit even before getting this far
```
bergarces, line 41: No, it shouldn't.

jpsains (approved):

jpsains (approved):

PR #433, opened 2025-05-15
observability changes

bergarces:


packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -108,16 +105,24 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
...
+      {
+        tokenAddresses,
+        currency: vsCurrency,
+        chainId: this.chainId,
+        url: urlObject.toString(),
+      },
+      'Fetching spot prices',
+    )
 
-      const response = await fetch(`${url}?${params.toString()}`, {
```
bergarces, line 120: It's safer to build the URL using the URL class (this would fail if querystring params were added to the base url)

packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -108,16 +105,24 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
...
+
+    logger.info(
+      {
+        tokenAddresses,
+        currency: vsCurrency,
+        chainId: this.chainId,
+        url: urlObject.toString(),
+      },
+      'Fetching spot prices',
+    )
```
bergarces, line 122: Added log every time we fetch spot prices, as this is not something we should be doing often and we can monitor if that's the case.

packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -201,6 +213,15 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
...
     }
 
+    logger.info(
+      {
+        tokenAddresses: tokenMetadata.address,
+        vsCurrency: USD,
+        chainId: this.chainId,
+      },
+      'Fetching on-chain prices',
+    )
```
bergarces, line 223: Log when we fetch on-chain prices (which should be never on deployed envs).

packages/adapters-library/src/core/utils/extractErrorMessage.ts
```diff
@@ -3,4 +3,4 @@ export const extractErrorMessage = (error: unknown): string =>
     ? error.message
     : typeof error === 'string'
       ? error
-      : 'Cannot extract error message'
+      : String(error)
```
bergarces, line 6: Better than a generic error.

jpsains (approved):

PR #434, opened 2025-05-15
better fetch for price-api

bergarces:


packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -129,18 +133,21 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
         },
       })
 
-      if (!response.ok) {
-        throw new Error(`Error fetching spot price: ${response.statusText}`)
+      if (!response.ok || response.status === 204) {
```
bergarces, line 136: 204  (No Content) is an ok response, but an empty body, which makes response.json() to throw. This treats it as an error.

packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -103,17 +105,19 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
...
 
   private async getSpotPriceByAddress({
-    tokenAddresses,
+    tokenAddress,
     vsCurrency = 'usd',
   }: GetSpotPriceByAddressInput): Promise<SpotPrice | null> {
-    const url = `${BASE_URL}/v1/chains/${this.chainId}/spot-prices/${tokenAddresses}`
+    const url = `${BASE_URL}/v1/chains/${this.chainId}/spot-prices/${
+      tokenAddress === E_ADDRESS ? ZERO_ADDRESS : tokenAddress
+    }`
```
bergarces, line 113: Just changing it at url level. The logs will display both the original token address and the url we are calling.

Prithpal-Sooriya (approved):

PR #437, opened 2025-05-21
cache and init improvements

bergarces:
- Refactor and simplify unwrap memory cache. It will no longer keep old cache entries.
- Removed some options from the API input (block numbers and token ids)
- Clean up some envs and config that were not being used.
- Remove adapters controller initialisation that we had to save some time when using lambdas


packages/adapters-library/src/core/unwrapCache.ts
bergarces, line 1: Moved to core and no longer takes a provider, since we are always using memory one.

If we decide to use something, like Redis, we can implement a new one using the `IUnwrapCache` interface and re-enable passing it at constructor time.

packages/adapters-library/src/core/unwrapCache.ts
```diff
@@ -0,0 +1,89 @@
...
+  fetchPriceWithCache(
+    adapter: IPricesAdapter,
+    input: PricesInput,
+  ): Promise<UnwrapExchangeRate>
+}
+
+export class MemoryUnwrapCache implements IUnwrapCache {
+  private readonly cache: Map<
+    string,
+    Promise<{ value: UnwrapExchangeRate; expiryTime: number }>
```
bergarces, line 25: Stores a promise, not the value itself, that way it prevents multiple requests for the same value when the first one is still resolving.

packages/adapters-library/src/core/unwrapCache.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  async fetchUnwrapWithCache(
+    adapter: IProtocolAdapter,
+    input: UnwrapInput,
+  ): Promise<UnwrapExchangeRate> {
+    return await this.fetchWithCache(
+      adapter.chainId,
+      input.protocolTokenAddress,
+      !!input.blockNumber,
```
bergarces, line 41: Passing a block number ignores the cache. We do not need to maintain block number caching as we no longer pass block numbers via API.

packages/adapters-library/src/config.ts
```diff
@@ -95,11 +95,6 @@ const ConfigSchema = z
       .default(
         parseBooleanEnv(process.env.DEFI_ADAPTERS_USE_MULTICALL_INTERCEPTOR),
       ),
-    useGetAllTransferLogs: z
-      .boolean()
-      .default(
-        parseBooleanEnv(process.env.DEFI_ADAPTERS_USE_GET_ALL_TRANSFER_LOGS),
-      ),
```
bergarces, line 102: This wasn't being used at all.

packages/adapters-library/src/adapters/prices-v2/products/usd/pricesV2UsdAdapter.ts
```diff
@@ -108,25 +109,25 @@ export class PricesV2UsdAdapter implements IPricesAdapter {
...
         chainId: this.chainId,
-        url: urlObject.toString(),
+        url: url.toString(),
       },
       'Fetching spot prices',
     )
 
     try {
-      const response = await fetch(urlObject.toString(), {
+      const response = await fetchWithRetry(url, {
```
bergarces, line 130: Uses this wrapper to retry once just in case something went wrong.

packages/adapters-api/src/schemas-request.ts
```diff
@@ -32,19 +32,17 @@ export const GetPositionsParamsSchema = z
...
   .object({
+    filterChainIds: parseAndCheck(z.array(z.nativeEnum(Chain)).optional()),
     filterProtocolIds: parseAndCheck(
       z.array(z.nativeEnum(Protocol)).optional(),
     ),
     filterProductIds: parseAndCheck(z.array(z.string()).optional()),
-    filterChainIds: parseAndCheck(z.array(z.nativeEnum(Chain)).optional()),
-    blockNumbers: parseAndCheck(z.record(z.string(), z.number()).optional()),
     filterProtocolTokens: parseAndCheck(z.array(z.string()).optional()),
-    filterTokenIds: parseAndCheck(z.array(z.string()).optional()),
```
bergarces, line 42: Remove option to pass block numbers or token ids to the API.

TokenIds are not being used (only morpho-blue, but still unclear why) and block numbers now ignore the cache and we don't want to support them via API for now.

packages/adapters-api/src/build-services.ts
bergarces, line 1: No need to pass an unwrap provider any longer.

packages/dev-ui/vite.config.ts
```diff
@@ -20,7 +20,6 @@ const envs = [
   'DEFI_ADAPTERS_PROVIDER_ARBITRUM',
   'DEFI_ADAPTERS_PROVIDER_AVALANCHE',
   'DEFI_ADAPTERS_PROVIDER_LINEA',
-  'DEFI_ADAPTERS_ENABLE_USD_PRICES_FOR_POSITIONS',
```
bergarces, line 23: This env does not exist.

packages/adapters-library/src/defiProvider.ts
```diff
@@ -139,10 +133,6 @@ export class DefiProvider {
   }): Promise<DefiPositionResponse[]> {
     const startTime = Date.now()
 
-    // TODO: We need to remove this. It can make the whole process crash if this function throws as it's not awaited
-    // It also does not need to be optimized for serverless
-    this.initAdapterControllerForUnwrapStage()
```
bergarces, line 144: This had to go. It was pointlessly being called on every use of getPositions.

packages/adapters-library/src/core/utils/fetchWithRetry.ts
bergarces, line 1: Simple retry handler set to do a single retry.

If we need to expand it to other parts of the code, we can have it take number of retries.

packages/adapters-api/src/schemas-request.ts
jpsains: agree we dont want to support that param - tokenIDs also used by uniswapV3 adapter

jpsains (dismissed):

PR #447, opened 2025-06-10
Worker tests refactor

bergarces:
image

### DIFF
--- a/packages/workers/src/cache/process-block.ts
+++ b/packages/workers/src/cache/process-block.ts
@@ -39,7 +39,7 @@ export async function processBlock({
     for (const log of receipt.logs || []) {
       const topic0 = log.topics[0]
       if (!topic0) {
-        // skip when no topic0 because our adapters expect a topic0 
+        // skip when no topic0 because our adapters expect a topic0
         continue
       }
 


packages/workers/src/api/build-api.ts
bergarces, line 1: No longer needs to call `getBlockNumber` during healthcheck, which seems to be what was causing restarts when it failed. It now gets reported from the main loop.

packages/workers/src/cache/build-latest-cache.ts
bergarces, line 1: Like history cache, this function now performs a single step of the loop, which is described in detail.

`waitForBlock` moved to its own file and improved so that:
- It is a single loop that never finishes until the next block is available
- All calls to the provider are caught.

'processEvent' also moved to its own file so it can be tested in isolation.

jpsains (approved):
You are a hero.

The tests are extreamly thorough - excelent work Mr B

jpsains (approved):

packages/workers/src/cache/process-block.ts
```diff
@@ -0,0 +1,80 @@
...
+      } catch (error) {
+        logger.error(
+          {
+            error: extractErrorMessage(error),
+            txHash: log.transactionHash,
+            eventAbi,
+            userAddressIndex,
+          },
+          'Error parsing log',
+        )
```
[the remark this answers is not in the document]
bergarces: If we remove the try/catch, then it will never be able to go pass that block and we would have a massive prod issue.

That `parseUserEventLog` has a few throws for things that should, in theory, never happen.

In an ideal world. we should never see that error in the logs (which I can confirm has never shown yet).

PR #452, opened 2025-06-12
feat: carbon user event

jpsains:
**Small changes**
* Added carbon user event 
* Added carbon protocol token

**Larger change** 

* Introduced openingPositionTxHash on the test case
* Added new test which is run if userEvent is defined on the adapter and if openingPositionTxHash is defined on test case

When the test is run we check that the open-Position-Transaction 1) has an event emitted from one of the protocol token contract and then 2) the event fully matches the defined event details in the adapter.

I found it useful - there are some leaps of faith but brought me confidence Ive defined the event correctly and it also reminded me that I had to define the protocolToken in the adapter which I nearly forgot.


Prithpal-Sooriya (approved):

packages/adapters-library/src/adapters/snapshots.test.ts
```diff
@@ -233,6 +234,141 @@ function runProductTests(
...
+          it.each(
+            positionsTestCases.map((testCase) => [testKey(testCase), testCase]),
+          )(
+            'User event for test %s match',
+            async (_, testCase) => {
+              const { defiProvider } = await fetchSnapshot(
+                testCase,
+                protocolId,
+                productId,
+              )
```
bergarces, line 260: nit, but we can use the checks to exit early instead of creating multiple levels of indentation?

Something like:

```
const adapter = defiProvider.adaptersController.fetchAdapter(
  testCase.chainId,
  protocolId,
  productId,
)

if (!testCase.input.openingPositionTxHash || !adapter?.adapterSettings.userEvent) {
  return
}
```

Happy to merge it anyway, we can polish this things later and this is not runtime code.

packages/adapters-library/src/adapters/snapshots.test.ts
```diff
@@ -1,9 +1,15 @@
 import assert from 'node:assert'
 import { promises as fs } from 'node:fs'
 import path from 'node:path'
+import {
+  buildTokenEventMappings,
+  createWatchKey,
+  processReceipts,
+} from '@metamask-institutional/workers'
```
jpsains, line 8: @bergarces I made some updates reusing functions from the worker library instead of writing bespoke code inside to validate userEvent set correctly but see red pipeline the snapshot tests failing to import. Ive tried a few things but no luck - did you already experiance something similar?

Prithpal-Sooriya (approved):

PR #460, opened 2025-07-15
feat: remove-transfer-log-fallback-code

jpsains:


packages/adapters-cli/src/commands/library-commands.ts
```diff
@@ -10,6 +11,9 @@ import type { Command } from 'commander'
 import { startRpcSnapshot } from '../utils/rpc-interceptor.js'
 import { extractRpcMetrics } from './build-scoreboard-command.js'
 
+const DEFAULT_USER_ADDRESS = '0x47ab2ba28c381011fa1f25417c4c2b2c0d5b4781'
```
bergarces, line 14: 🙇

packages/adapters-library/src/defiProvider.ts
```diff
@@ -131,6 +123,26 @@ export class DefiProvider {
     filterProtocolTokens?: string[]
     filterTokenIds?: string[]
   }): Promise<DefiPositionResponse[]> {
+    console.log({
```
bergarces, line 126: No `console.log` in library code.

packages/adapters-cli/src/commands/library-commands.ts
bergarces, line 14: Any reason to have a default user address? This can be annoying if such address position changes over time.

packages/adapters-cli/src/commands/library-commands.ts
```diff
@@ -40,11 +44,56 @@ export function libraryCommands(program: Command) {
         const filterProductIds = multiProductFilter(productIds)
         const filterChainIds = multiChainFilter(chains)
 
+        const resolvedUserAddress = userAddress ?? DEFAULT_USER_ADDRESS
+        if (!userAddress) {
```
bergarces, line 48: When is this condition going to trigger? `DEFAULT_USER_ADDRESS` is a non-empty string.

packages/adapters-cli/src/commands/library-commands.ts
```diff
@@ -40,11 +44,56 @@ export function libraryCommands(program: Command) {
...
+          console.log(
+            `No user address provided. Using default address ${DEFAULT_USER_ADDRESS}`,
+          )
+        }
+
         const filterProtocolTokens =
           multiProtocolTokenAddressFilter(protocolTokens)
 
+        if (!buildPoolFilter() && !filterProtocolTokens) {
```
bergarces, line 57: Only call `buildPoolFilter()` once, otherwise it's creating a bunch of pools each time.

packages/adapters-cli/src/commands/library-commands.ts
jpsains: If you run npm positions without a user address it passes userAddress = undefined to the adapters the error is awful - Ive updated to use the default feature on commander

PR #472, opened 2025-09-05
feat: metadata storage support

jpsains:
## **Description**

This PR enables tracking things like multiple public keys for ETH2 staking and multiple token IDs for Uniswap V4

## Problem Statement
Previously, the system could only store one-to-one relationships between users and contract addresses. This limitation prevented:
- **ETH2 Staking**: Users with multiple validator public keys linked to the same withdrawal address
- **Uniswap V4**: Users with multiple token IDs for the same contract
- **Future Protocols**: Any protocol requiring multiple metadata values per user-contract pair

### 1. Database Schema Changes
- **Added metadata columns** to `logs` table: `metadata_key` and `metadata_value` (TEXT)
- **Updated unique constraint** from `PRIMARY KEY (address, contract_address)` to `UNIQUE (address, contract_address, metadata_key, metadata_value)`
- **Added job configuration** columns: `additional_metadata_arguments` (JSONB) and `transform_user_address_type` (TEXT)

### 2. Type System Updates
- **Renamed `PoolFilter`** to `DefiPositionDetection` for better clarity
- **Extended `AdapterSettings`** to support metadata extraction and address transformation
- **Added `UserAddressTransformer`** type for reusable address transformation functions

### 3. Event Processing Enhancements
- **Generic metadata extraction** from event logs using configurable argument mappings
- **Address transformation system** for handling non-standard address formats (e.g., ETH2 withdrawal credentials)
- **Graceful handling** of unsupported address types (e.g., BLS credentials)

### Metadata Storage
- **Flexible metadata keys**: Support for any string-based metadata identifier
- **Multiple values per user-contract pair**: Enables 1:many relationships
- **Efficient querying**: Optimized indexes for fast metadata retrieval

### Address Transformation
- **Reusable transformers**: Generic system for converting raw event data to addresses
- **ETH2 support**: Handles withdrawal credentials transformation (`0x01` → address, `0x00` → skip)
- **Extensible design**: Easy to add new transformation types

### Event Configuration
- **Declarative metadata mapping**: Adapters specify which event arguments to extract as metadata
- **Type-safe configuration**: Compile-time validation of event ABI and argument names
- **Backward compatibility**: Existing adapters continue to work without changes

### Database Migrations
1. **0003-update-metadata-to-arrays.ts**: Adds metadata columns and job configuration and updates unique constraint to support multiple metadata entries

### Core Components
- **`parseUserEventLog`**: Enhanced to extract metadata and transform addresses
- **`UserAddressTransformer`**: Reusable transformation functions
- **`DefiPositionDetection`**: Updated to return both contract addresses and metadata
- **`DefiProvider`**: Integrates metadata into adapter calls

### Adapter Updates
- **ETH2 Validator Staking**: Configured to track `DepositEvent` with public key metadata
- **Uniswap V4**: Example configuration for `Transfer` events with token ID metadata

### Before
```sql
CREATE TABLE logs (
  address TEXT,
  contract_address TEXT,
  PRIMARY KEY (address, contract_address)
);
```

### After
```sql
CREATE TABLE logs (
  address TEXT,
  contract_address TEXT,
  metadata_key TEXT,
  metadata_value TEXT,
  UNIQUE (address, contract_address, metadata_key, metadata_value)
);
```

## Conclusion
This PR provides a robust, scalable solution for storing and retrieving metadata associated with DeFi positions. The implementation maintains backward compatibility while enabling new use cases that require 1:many relationships between users and their DeFi assets.

This foundation will support current requirements (ETH2 staking, Uniswap V4) while providing flexibility for future DeFi protocols that require similar metadata tracking capabilities.

---

> [!NOTE]
> Adds metadata-aware DeFi position detection and storage (e.g., ETH2 pubkeys, Uniswap V4 tokenIds) with DB/schema changes, workers/library updates, and API/CLI integration.
> 
> - **Database/Schema**:
>   - Add `logs` metadata columns: `metadata_key`, `metadata_value`; make row uniqueness `UNIQUE(address, contract_address, metadata_key, metadata_value)`.
>   - Extend `jobs` with `additional_metadata_arguments` and `transform_user_address_type`.
>   - New migration `0003-update-metadata-to-arrays`.
> - **Workers**:
>   - Parse events with optional address transformers and extract metadata; propagate via `UserIndex`.
>   - Write metadata-aware logs; update historic/latest cache builders and processing pipeline.
>   - Job selection now carries metadata config; PostgreSQL client supports metadata insert/query.
> - **Library/Core**:
>   - Replace `PoolFilter` with `DefiPositionDetection` returning contract addresses plus metadata per contract.
>   - `DefiProvider` consumes detection results and forwards combined metadata as `tokenIds` to adapters.
>   - Add `LOG_ARGUMENT_TRANSFORMERS` (e.g., `eth2-withdrawal-credentials`).
>   - Pass `config` through controllers/helpers; expose new types/exports.
> - **Adapters**:
>   - ETH2 staking: configure `userEvent` (DepositEvent) with pubkey metadata; use configurable beacon API.
>   - Uniswap V4: enable `Transfer`-based `userEvent` and extract `tokenId` metadata.
> - **API/CLI**:
>   - API `/user-pools/:userAddress` and services use `getDefiPositionsDetection*`.
>   - CLI switches to PostgreSQL-backed detection and adds setup guidance/connection checks.
> - **Config/Docs/Tests**:
>   - Add beacon node envs; update rules/docs for metadata.
>   - Update tests and snapshots for metadata flow and new APIs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c600f90dd4892b5077d0dcccf23a411a16445f5a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/workers/src/cache/get-next-pool-group.ts
```diff
@@ -64,6 +64,8 @@ export async function getNextPoolGroup(
       topic0: largestGroup[0]!.topic0,
       eventAbi: largestGroup[0]!.eventAbi,
       userAddressIndex: largestGroup[0]!.userAddressIndex,
+      additionalMetadataArguments: largestGroup[0]!.additionalMetadataArguments,
+      transformUserAddressType: largestGroup[0]!.transformUserAddressType,
```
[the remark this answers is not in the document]
bergarces: That's needed from time to time when `noUncheckedIndexedAccess: true`. It's an statement from the developer stating that they have checked there's an entry beforehand.

It's significantly unsafer when `noUncheckedIndexedAccess: false` and typescript just assumes the entry is not undefined, as there's no warning.

packages/workers/src/cache/user-address-transformers.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+export function getUserAddressTransformer(
+  transformType?: string,
+): UserAddressTransformer | undefined {
+  if (!transformType) {
+    return undefined
+  }
+
+  return USER_ADDRESS_TRANSFORMERS[transformType]
```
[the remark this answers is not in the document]
jpsains: hey, 
we have     "noUncheckedIndexedAccess": true set 

so get 
Type 'undefined' cannot be used as an index type 

but ill simplify to 

```transformType ? USER_ADDRESS_TRANSFORMERS[transformType] : undefined```

File: packages/adapters-api/src/db-service.ts [hunk 3]
```diff
@@ -57,33 +71,64 @@ export class PostgresService implements DbService {
     this.#dbPools = dbPools
   }
 
-  async getAddressPools(
+  async getDefiPositionsDetectionPerChain(
     userAddress: string,
   ): Promise<Partial<Record<EvmChain, string[]>>> {
     return Object.values(EvmChain).reduce(
```
bergarces, line 77: Nothing to do with this PR, but I've just realised that this construct will run the loop of `getAddressChainPools` sequentially, not in parallel.

It's probably not worth worrying about it, but if that is the case maybe we should get the results for each chain as a map with a `Promise.all` and then build the object.

packages/workers/src/database/migrations/0003-update-metadata-to-arrays.ts
```diff
@@ -0,0 +1,37 @@
...
+  // Add transformUserAddressType column to jobs table
+  await client.query(
+    'ALTER TABLE jobs ADD COLUMN IF NOT EXISTS transform_user_address_type TEXT',
+  )
+
+  // Remove the existing primary key constraint on (address, contract_address)
+  // This allows multiple metadata entries per user-contract pair
+  await client.query(`
+    ALTER TABLE logs DROP CONSTRAINT IF EXISTS logs_pkey
+  `)
```
bergarces, line 30: I don't know the exact details of how postgres deals with indexes. I'm pretty certain that creating a primary key would have created an index.

If this is removed, I'd keep a keen eye on how the query, which filters by `address` (user address) performs. If it performs worse than before, you might need to manually add a filter for just that column.

bergarces (dismissed):

packages/adapters-cli/src/commands/library-commands.ts
```diff
@@ -53,17 +59,29 @@ export function libraryCommands(program: Command) {
...
         }
 
-        const filter: PoolFilter = async (userAddress, chainId) => {
+        const getDefiPositionsDetection: DefiPositionDetection = async (
+          userAddress,
+          chainId,
+        ) => {
           if (filterUsingLocalIndex) {
-            return buildFilterUsingLocalIndex()!(userAddress, chainId)
+            return await filterUsingLocalIndex(userAddress, chainId)
```
bergarces, line 84: Are you sure this needs to be awaited?

